PR #96, opened 2015-06-22
Fix for a problem with active attribute updates coming from Context Broker

marcinbudny:
1) The IoT agent crashes when active attribute update arrives from Context Broker and `commands` property of the entity type is not defined. Problem lays in `contextServer.executeUpdateSideEffects` not checking if the `commands` property is defined.
2) Once that was fixed, it turned out wrong attributes were passed to client handler. In `contextServer.generateUpdateActions` they were taken from `device` instead of `contextElement`.

### DIFF
--- a/lib/services/contextServer.js
+++ b/lib/services/contextServer.js
@@ -171,29 +171,32 @@ function createQueryResponse(req, res, results) {
 function executeUpdateSideEffects(device, id, type, service, subservice, attributes, callback) {
     var sideEffects = [];
 
-    for (var i = 0; i < device.commands.length; i++) {
-        for (var j = 0; j < attributes.length; j++) {
-            if (device.commands[i].name === attributes[j].name) {
-                var newAttributes = [
-                    {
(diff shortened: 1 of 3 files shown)


dmoranj:
Sorry for the delay attending your PR. A couple of things: first of all, as I commented in another PR, contributions should always target the `develop` branch (even in cases like this one, when both branches have currently the same content). So, please, close this PR and create a new one from your `master` to `develop`. For future contributions, you should work in your `develop` branch as well, as that one will differ from `master` from time to time. I have added some information in the documentation about the releasing process (in the `develop` branch), that I hope will clarify how the branching and releasing processes work.

Another important thing to do before contributing is to execute:
```
grunt
```
on the root folder of your branch, to check that both all the tests and syntax rules comply with those of the project. In this case, I have downloaded your branch and some syntax rules are broken. 

So, please, check the syntax problems in your branch and create a new PR targeting `develop`. And thanks for your contribution.

PR #108, opened 2015-07-21
Fix/memory registry store error

dmoranj:
Fix callback in the `store()` function of the Memory Registry (didn't return the device).


dmoranj:
Close PR to the wrong branch. Again. :@

PR #127, opened 2015-08-14
Bug/return cb error details

### LINKED ISSUE #125: Information of an error response from the Context Broker is not propagated
When a function of the library is executed and the Context Broker answers with a non-200 statusCode, a BAD_REQUEST error is returned, always with code 400 and a standard text. It should return the code indicated in the Orion error and the details returned by the CB.

dmoranj:
Fixes #125

Add error information to the generic NGSI error.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -32,9 +32,9 @@ module.exports = {
         this.name = 'UNREGISTRATION_ERROR';
         this.message = 'Error unregistering context provider for device: ' + id + ' of type: ' + type;
     },
-    EntityUpdateError: function(id, type, details) {
-        this.name = 'ENTITY_UPDATE_ERROR';
-        this.message = 'Error updating entity data for device: ' + id + ' of type: ' + type;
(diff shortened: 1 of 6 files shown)


lib/errors.js
```diff
@@ -32,9 +32,10 @@ module.exports = {
         this.name = 'UNREGISTRATION_ERROR';
         this.message = 'Error unregistering context provider for device: ' + id + ' of type: ' + type;
     },
-    EntityUpdateError: function(id, type) {
-        this.name = 'ENTITY_UPDATE_ERROR';
-        this.message = 'Error updating entity data for device: ' + id + ' of type: ' + type;
+    EntityGenericError: function(id, type, details) {
+        this.name = 'ENTITY_GENERIC_ERROR';
+        this.message = 'Error accesing entity data for device: ' + id + ' of type: ' + type;
+        this.details = details;
```
mrutid: Take into account when detail is null or undefined. Assign empty object instead.

mrutid:
OK

PR #171, opened 2015-10-20
FIX Use express methods to check mime type

### LINKED ISSUE #170: Use Express methods to check the content type
The content type is being currently tested checking the literal value of the headers. The `is()` method of the Express request should be used instead.

dmoranj:
Fixes #170


XavierVal:
LGTM

PR #179, opened 2015-11-07
Fix bug when trying to stop an unstarted agent

LeandroGuillen:
agentConsole.js crashed violently when issuing a *stop* command and no *start* command had been performed.

### DIFF
--- a/lib/services/northBound/northboundServer.js
+++ b/lib/services/northBound/northboundServer.js
@@ -117,7 +117,9 @@ function start(config, callback) {
 function stop(callback) {
     logger.info(context, 'Stopping IoT Agent');
 
-    northboundServer.server.close(callback);
+    if (northboundServer) {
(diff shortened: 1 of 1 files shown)


lib/services/northBound/northboundServer.js
```diff
@@ -117,7 +117,9 @@ function start(config, callback) {
 function stop(callback) {
     logger.info(context, 'Stopping IoT Agent');
 
-    northboundServer.server.close(callback);
+    if (northboundServer) {
+        northboundServer.server.close(callback);
+    }
```
dmoranj: In case it's stopped, the callback must be called, anyways, or the IOTA will get stuck.

dmoranj:
I have just noticed the PR si pointing to master. It shouldn't. Fixes must be merged with develop, only releases merge with master. I'll close this PR and open another one from develop with the same fix.

PR #181, opened 2015-11-10
FIX Allow list devices to use a single param

### LINKED ISSUE #180: ListDevices doesn't work with just one parameter
The current listDevices implementation for both device registries (in-memory and MongoDB) works only in the four parameters form, and is should work with just the callback (to return the complete collection of devices). The in-memory registry has an additional problem, as it forces the existence of the service and subservice parameters (that should be optional).

dmoranj:
Fixes #180


CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Allow listDevices() to use a single param (#180).
```
gtorodelvalle: Final blank line appreciated... :)

AlvaroVega:
LGTM

lib/services/devices/deviceService.js
```diff
@@ -403,7 +403,15 @@ function updateRegisterDevice(deviceObj, callback) {
...
-        callback = limit;
+        if (service && subservice && limit) {
+            callback = limit;
+        } else if (service) {
+            callback = service;
+            service = null;
+            subservice = null;
+        } else {
+            logger.fatal(context, 'Couldn\'t find callback in listDevices() call.');
+        }
```
gtorodelvalle: Kind of confusing to me these assignments to the `callback` parameter... :) Taking into consideration it ends up being passed to the listDevices() function at https://github.com/telefonicaid/iotagent-node-lib/blob/bb39fff6494df14437c56b85e1b7cf8e7b8f8039/lib/services/devices/deviceRegistryMongoDB.js#L107 Maybe it is because it is Tuesday after a long weekend... :)

gtorodelvalle:
LGTM and merging! ;) Thanks @dmoranj :)

PR #190, opened 2015-11-26
FIX Add service and subservice headers in remove device

dmoranj:
Remove device operation in the IoT Agent client won't work through authenticated PEP Proxies, as no fiware-service and fiware-servicepath headers were sent.


XavierVal:
LGTM

XavierVal:
LGTM

PR #191, opened 2015-11-26
Bug/expose find group functionality

dmoranj:
Expose the find group features (existing in the groupRegistry module, but not exposed publicly in the library object).


XavierVal:
LGTM

PR #192, opened 2015-11-27
Task/include groups in clear all

dmoranj:
Clear also the Group registry (not only the Device registry) in the library's clearAll() function.


XavierVal:
LGTM

PR #197, opened 2015-12-02
ADD Prefix /iot to the IoTAgent registration in the IOTAM

dmoranj:
Add /iot prefix for IoTAM integartion.

### DIFF
--- a/lib/services/northBound/deviceProvisioningServer.js
+++ b/lib/services/northBound/deviceProvisioningServer.js
@@ -127,7 +127,10 @@ function handleListDevices(req, res, next) {
             if (error) {
                 next(error);
             } else {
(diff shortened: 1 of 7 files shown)


lib/services/northBound/deviceProvisioningServer.js
```diff
@@ -127,7 +127,10 @@ function handleListDevices(req, res, next) {
             if (error) {
                 next(error);
             } else {
-                res.status(200).json(deviceList);
+                res.status(200).json({
+                    count: deviceList.length,
```
anabelengp: API clients need the total count of registered devices for pagination purposes

anabelengp:
LGTM

PR #201, opened 2015-12-10
Task/append mode

dmoranj:
Add an APPEND mode to send APPEND requests instead of UPDATE ones whenever an update function is called.


XavierVal:
LGTM

PR #208, opened 2015-12-16
Refactor the time compression plugin to extract common features

### LINKED ISSUE #205: Refactor the generic functions used in the time compression plugin to a common module
Most of the code of the time compression plugin is composed of functions that act apply generic functions into different parts of the entity. This code seems aplicable to lots of other possible plugins, so it should be refactored to a common module for its use.

dmoranj:
Fixes #205


XavierVal:
LGTM

PR #212, opened 2015-12-22
FIX Queries to the registries based on service

dmoranj:
Queries to the MongoDB registries based on service was not working (the condition was stablished after the query was executed).


AlvaroVega:
LGTM

XavierVal:
LGTM

PR #221, opened 2015-12-23
FIX Extra parameter in query callback

### LINKED ISSUE #220: Query middlewares won't work
Due to the last changes in the data transformation plugins, the `ngsi.query` callback received an extra parameter, making queries with query middlewares fail.

dmoranj:
Fixes #220


anabelengp:
LGTM

PR #226, opened 2016-01-07
Task/extract command line handlers

### LINKED ISSUE #178: Extract the command line handlers to a single module
The IOTA command line tester executes some commands that may be useful for the testers of specific IOTAs, but they are currently unusable, as they are encapsulated in the IOTA Tester module. They should be extracted to a different independent module, that could that way be reused in different specific IOTA testers for free.

dmoranj:
Fixes #178


XavierVal:
LGTM

PR #231, opened 2016-01-13
Task/add global error handler

### LINKED ISSUE #229: Add a global error handler
Add a global error handler that prevents the agent from crashing upon unexpected exceptions.

(There is some offline debate about whether to add this feature or not, but it will be added in a configurable manner).

dmoranj:
Fixes #229


XavierVal:
LGTM

PR #236, opened 2016-01-13
FIX Appropriate name for attribute ids

### LINKED ISSUE #233: Attribute id in alias should be named 'object_id'
In the current implementation of the attribute alias mechanisms, the id of the device attribute is named "id", but it should be named "object_id".

dmoranj:
Fixes #233


XavierVal:
LGTM

PR #241, opened 2016-01-21
Fix/internal error code leads to wrong http code

### LINKED ISSUE #239: Library errors make HTTP Response invalid
When a library returns an error object with a .code attribute, the contained value is returned, even if its not a valid HTTP code.

dmoranj:
Fixes #239


gtorodelvalle:
LGTM and merging... ;) Thanks!

PR #246, opened 2016-01-27
FIX Object IDs in get provisioned device

### LINKED ISSUE #245: Getting the information of a provisioned device returns an spurious 'id' attribute in attribute descriptions
A little back this year, the name of the attribute's attribute, 'id' was changed to 'object_id'. It seems part of the code remains as it was in the past, so the 'id' attribute is coming in the get provisioned device operation. It should be removed.

dmoranj:
Fixes #245


AlvaroVega:
LGTM

PR #253, opened 2016-01-28
FIX Both object_id and id stored for attributes

### LINKED ISSUE #250: Both object_id and id in attributes are stored, but only the id displayed
When the object_id was introduced (as the external name for the attribute id), id was kept as the internal name for the id. This caused a duplication of the attribute in the database that cause some visualization problems. The best option may be to change the internal code to use object_id for everyting.

dmoranj:
Fix #250


XavierVal:
LGTM

PR #266, opened 2016-02-10
FIX Change timestamp type to 'ISO8601'

### LINKED ISSUE #265: Change timestamp type to 'ISO8601'
Current timestamp's attribute type is "timestamp". It should be changed to "ISO8601".

dmoranj:
Fixes #265

### DIFF
--- a/test/unit/compress-timestamp-plugin_test.js
+++ b/test/unit/compress-timestamp-plugin_test.js
@@ -146,7 +146,7 @@ describe('Timestamp compression plugin', function() {
             },
             {
                 name: 'The Target Value',
-                type: 'timestamp',
+                type: 'ISO8601',
                 value: '20071103T131805'
             }
         ];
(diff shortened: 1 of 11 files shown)


test/unit/compress-timestamp-plugin_test.js
```diff
@@ -146,7 +146,7 @@ describe('Timestamp compression plugin', function() {
             },
             {
                 name: 'The Target Value',
-                type: 'timestamp',
+                type: 'ISO8601',
                 value: '20071103T131805'
```
AlvaroVega: is this value a right time ISO8601 ?

test/unit/contextRequests/updateContextCompressTimestamp1.json
```diff
@@ -12,7 +12,7 @@
         },
         {
           "name": "The Target Value",
-          "type": "timestamp",
+          "type": "ISO8601",
           "value": "+002007-11-03T13:18:05"
```
AlvaroVega: I miss Timezone in value:```+002007-11-03T13:18:05Z```

test/unit/compress-timestamp-plugin_test.js
dmoranj: Quoting from Wikipedia:
```
Alternatively, a format for duration based on combined date and time representations may be used by agreement between the communicating parties either in the basic format PYYYYMMDDThhmmss or in the extended format P[YYYY]-[MM]-[DD]T[hh]:[mm]:[ss].
```
Maybe the P is missing is that's what your refer to. This format was the compressed one agreed with the Thinking Things for its use inside the devices. The IoTAgent itself transforms it to the expanded representation using the time compression plugin.

AlvaroVega:
LGTM

PR #282, opened 2016-02-22
Task/refactor configuration inits

### LINKED ISSUE #167: Create a config service to avoid multiple initializations
Most of the modules require the global configuration object, and it is currently being inserted using an `init()` method in each module, and storing it separately. This should be changed to a config service that could be asked for a value when needed, in order to avoid the need of initialization of every module in the main server.

### LINKED ISSUE #173: Use a single database configuration for all persistence services
Currently, for each persistence service (Devices, Groups, Stats), there is a different database configuration. This adds to the flexibility of the library but is a real pain regarding configuration (and it doesn't seems to be a really useful flexibility feature). All these configurations should be colapsed in a single configuration.

dmoranj:
Fixes #167


lib/services/devices/deviceService.js
```diff
@@ -27,15 +27,14 @@ var request = require('request'),
...
     subscriptions = require('../ngsi/subscriptionService'),
     _ = require('underscore'),
     context = {
         op: 'IoTAgentNGSI.DeviceService'
-    },
-    registry,
-    groupRegistry,
-    config;
+    }
+    ;
```
gtorodelvalle: I know it is kind of absurd this comment, but maybe you would like to move the `;` to the previous line... :)

gtorodelvalle:
Minor comment included but LGTM! ;) Thanks @dmoranj :)

PR #283, opened 2016-02-22
Task/single database configuration

### LINKED ISSUE #173: Use a single database configuration for all persistence services

dmoranj:
Fixes #173


AlvaroVega:
LGTM

Issue #263, opened 2016-02-09
Initial entity provisioning does not take configurations into account


dmoranj:
When a new device is provisioned, its representing entity is created in the Context Broker with initial values, to allow all the updateContext operations to use the UPDATE action type (that requires the presence of the attribute that is going to be modified). In the case where the attributes are defined at Configuration level (not device level), this creation fails, as it uses just the attributes defined in the provisioned device (instead of using the merge of both). Similar problems may be found for different scenarios for the same causes: the multiplicity of sources of definition for the devices configuration. We should think carefully about how to solve this problem as we may solve future bugs if we treat it carefully (maybe moving around an object that is a merge of both the device information and the configuration one, making the former override the latter).

Issue #260, opened 2016-02-05
Entity retrieval on incoming notifications or queries uses the entity name without the type


dmoranj:
When a request comes to the IOTA Northbound interface, the IOTA uses the entity name alone to retrieve the information of the device that generated the entity. This is an error, as multiple entities with the same name but different type may coexist for a single service. The identification should be changed to take this into account.

A similar problem may exist for entities with the same name in different services.

Issue #17, opened 2015-01-21
Change the Identification of the Devices.


dmoranj:
Early in the development of the IoT Agent, the library API was thought to use the combination of deviceId and deviceType as a key to distinguish different devices. This assumption doesn't hold anymore, so it's time to change it.

dmoranj:
The right identification pair is (Service, DeviceID).

PR #291, opened 2016-03-31
Task/rationalize service data

dmoranj:
This PR makes some refactors on the deviceService.js file, to make it more readable and modifiable, in order to prepare for important changes suggested in bugs #263 and #260 and the old feature #17.


gtorodelvalle:
LGTM

PR #296, opened 2016-04-18
ADD Endpoint attribute to Device model

dmoranj:
Add endpoint attribute to Device Provisioning API and device model.


AlvaroVega:
LGTM

Issue #303, opened 2016-04-26
Add support for MongoDB Replica Sets


dmoranj:
The mongodb driver should be able to connect to a MongoDB replicaset. In order to configure the replicaset:
- The `host` attribute of the mongodb configuration will accept multiple IPs sepparated by comma.
- A new attribute `replicaSet` will appear to give the driver the name of the replica set.

PR #305, opened 2016-04-26
Task/add replica set configuration

dmoranj:
Implements #303

### DIFF
--- a/lib/model/dbConn.js
+++ b/lib/model/dbConn.js
@@ -47,10 +47,18 @@ function loadModels() {
  *
  * @this Reference to the dbConn module itself.
  */
-function init(host, db, port, options, callback) {
+function init(host, db, port, username, password, options, callback) {
     /*jshint camelcase:false, validthis:true */
+    var url,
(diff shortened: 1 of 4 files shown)


lib/model/dbConn.js
```diff
@@ -47,10 +47,37 @@ function loadModels() {
...
+        previous += current;
+
+        return previous;
+    }
+
+    hosts = host.split(',')
+        .map(addPort)
+        .reduce(commaConcat, '');
+
+    url = 'mongodb://' + credentials + hosts + ':' + port + '/' + db;
```
gtorodelvalle: I think the `+ ':' + port` is not needed since this way you may end up with a connection URL such as: `mongodb://gtv:hey@mongodb1:12345,mongodb2:12345**:12345**/database`

dmoranj: Completely right. Funny thing is, I tested this patch in pre environment and it worked. Fixing it...

gtorodelvalle:
LGTM

PR #311, opened 2016-04-28
ADD Exported command line library

dmoranj:
Export command-line library for IOTAs using this library, to avoid the iotas importing a different version of the same command-line library.


anabelengp:
LGTM

PR #314, opened 2016-04-29
FIX Static attributes in query context

### LINKED ISSUE #312: When a query comes to the IOTA the Static Attributes should not be returned
The current behavior of the IOTA is to return the set of static attributes in every NGSI request, including queries; but this behavior only makes sense in updates. The static attributes should be removed from queries.

dmoranj:
Fixes #312


gtorodelvalle:
LGTM

PR #321, opened 2016-05-03
Task/integrate groups

### LINKED ISSUE #318: Configuration API Should allow to modify API Key
The API Key was originally conceived to be part of the Key for Configurations, but that criteria has changed, and now it should be allowed to change.

### LINKED ISSUE #320: Timestamp and level missing in logs
At some point during the development of the current version, timestamp and log level ceased to be printed.

dmoranj:
Fixes #320 
Fixes #318 

Fixes multiple problems found during integration with the Administration Portal.

### DIFF
--- a/lib/model/Group.js
+++ b/lib/model/Group.js
@@ -38,7 +38,7 @@ var Group = new Schema({
     timezone: String,
     commands: Array,
     lazy: Array,
-    active: Array,
+    attributes: Array,
     internalAttributes: Array
 });
 
--- a/lib/model/Group.js
+++ b/lib/model/Group.js
@@ -26,7 +26,6 @@ var mongoose = require('mongoose'),
     Schema = mongoose.Schema;
 
 var Group = new Schema({
-    id: String,
     url: String,
     resource: String,
     apikey: String,
--- a/lib/services/groups/groupRegistryMongoDB.js
+++ b/lib/services/groups/groupRegistryMongoDB.js
@@ -38,8 +38,8 @@ var logger = require('logops'),
  *
  * @return {Function}       The generated handler.
  */
-function saveGroupHandler(callback) {
-    return function saveHandler(error, groupDAO) {
+function saveGroupHandler(groupDAO, callback) {
+    return function saveHandler(error, result) {
         if (error) {
             logger.debug(context, 'Error storing device group information: %s', error);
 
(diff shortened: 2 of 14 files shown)


lib/model/Group.js
```diff
@@ -38,7 +38,7 @@ var Group = new Schema({
     timezone: String,
     commands: Array,
     lazy: Array,
-    active: Array,
+    attributes: Array,
     internalAttributes: Array
 });
```
gtorodelvalle: Since I saw you are using the default `_id` property for the stored (`Group`) document identifiers, I think you can remove the `id` property from the `Group` schema, can't you? Just to avoid confusion in the future... :) See: https://github.com/telefonicaid/iotagent-node-lib/blob/task/integrateGroups/lib/model/Group.js#L29

dmoranj: Completely right.

dmoranj: Fixed in 			3bf797f

lib/services/groups/groupRegistryMongoDB.js
```diff
@@ -65,7 +65,7 @@ function createGroup(group, callback) {
             'timezone',
             'commands',
             'lazy',
-            'active',
+            'attributes',
```
gtorodelvalle: Should the `id` property be `_id` instead (since I think you decided to use the standard `_id` property)? :)

lib/services/northBound/deviceGroupAdministrationServer.js
```diff
@@ -67,7 +67,11 @@ function handleCreateDeviceGroup(req, res, next) {
     for (var i = 0; i < req.body.services.length; i++) {
         req.body.services[i].service = req.headers['fiware-service'];
         req.body.services[i].subservice = req.headers['fiware-servicepath'];
+        req.body.services[i].type = req.body.services[i]['entity_type'];
         req.body.services[i].internalAttributes = req.body.services[i]['internal_attributes'];
+
+        //delete req.body.services[i]['entity_type'];
+        //delete req.body.services[i]['internal_attributes'];
```
gtorodelvalle: Not really sure if you would like to keep the previous 2 commented lines... :)

dmoranj: I didn't. I meant to add them, but I remove them to test a couple of things. I'll uncomment them now.

gtorodelvalle:
LGTM :)

PR #337, opened 2016-05-11
FIX Multiple timestamp additions

### LINKED ISSUE #335: Timestamp addition for static attributes generate multiple identical timestamp attributes
When a device defines static_attributes, and those attributes are sent to the Context Broker, the Timestamp is also added to the attributes. But when a single device request generates multiple NGSI updateContext, the Timestamp attribute in the metadatas is multiplied times the number of requests. This seems to be a problem with metadata addition in the DB (the same object retrieved once from the database is modified instead of creating a new one, and the timestamp is added instead of modified).

dmoranj:
Fixes #335

### DIFF
--- a/lib/services/ngsi/ngsiService.js
+++ b/lib/services/ngsi/ngsiService.js
@@ -195,6 +195,7 @@ function addTimestamp(payload) {
                 attribute.metadatas[i].name === constants.TIMESTAMP_ATTRIBUTE) {
                 attribute.metadatas[i].value = timestamp.value;
                 timestampFound = true;
+                break;
             }
         }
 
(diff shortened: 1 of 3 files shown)


lib/services/ngsi/ngsiService.js
```diff
@@ -184,11 +184,23 @@ function addTimestamp(payload) {
...
         if (!attribute.metadatas) {
             attribute.metadatas = [];
         }
 
-        attribute.metadatas.push(timestamp);
+        for (var i = 0; i < attribute.metadatas.length; i++) {
+            if (attribute.metadatas[i].type === constants.TIMESTAMP_TYPE &&
+                attribute.metadatas[i].name === constants.TIMESTAMP_ATTRIBUTE) {
+                attribute.metadatas[i].value = timestamp.value;
+                timestampFound = true;
```
gtorodelvalle: You could `break` the loop to avoid further processing once the `timestamp` metadata entry is found and its value updated :)

gtorodelvalle:
LGTM and merging once Travis is OK with it... ;)

Issue #341, opened 2016-05-12
Refactor the tests in order to ease the creation of new ones


dmoranj:
With the current organization of the test files (all in the same folder, along with the exmaple folders), it's pretty difficult to find where is the test to modify to add cases for a particular issue. They should be moved to folders per category, with a single folder with all the example folders.

PR #342, opened 2016-05-12
Task/reorganize tests

### LINKED ISSUE #340: When persistence is defined for stats but not for the registries, the IOTA crashes
When the stats service is configured with the peristence option on, and the mongodb data is configured, if the deviceRegistry options is not configured with MongoDB persistence as well, the agent doesn't initialize the db and crashes.

dmoranj:
Implements #341
Fixes #340


XavierVal:
LGTM

PR #344, opened 2016-05-13
Task/command updates not implemented

### LINKED ISSUE #308: Command updates not implemented
Currently, the device provisioning update do not accept commands, and silently ignore it. Commands should be modifiable in updates, creating the appropriate registrations and collateral attributes.

dmoranj:
Fixes #308

### DIFF
--- a/lib/services/devices/deviceService.js
+++ b/lib/services/devices/deviceService.js
@@ -326,6 +326,7 @@ function updateRegisterDevice(deviceObj, callback) {
         if (oldDevice) {
             oldDevice.internalId = newDevice.internalId;
             oldDevice.lazy = newDevice.lazy;
+            oldDevice.commands = newDevice.commands;
             oldDevice.active = newDevice.active;
             oldDevice.name = newDevice.name;
             oldDevice.type = newDevice.type;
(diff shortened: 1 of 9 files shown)


lib/services/devices/deviceService.js
```diff
@@ -337,36 +338,44 @@ function updateRegisterDevice(deviceObj, callback) {
...
-                id: oldDevice.id,
-                name: oldDevice.name,
-                type: oldDevice.type,
-                service: oldDevice.service,
-                subservice: oldDevice.subservice,
-                commands: [],
-                staticAttributes: []
-            },
-            oldActiveKeys,
+    function mixAttributeArrays(oldArray, newArray) {
```
gtorodelvalle: I am not sure if this is the best name for this function taking into consideration what it does... :) Shouldn't it be something such as `getNewAttributes()` since it returns attributes included in `newArray` but not included in `oldArray` (if I am not wrong :))

dmoranj: Yeah, maybe it's not the best name. GetAttributeDifference may be more appropriate.

gtorodelvalle: I didn't opted for one like that since there may be attributes in `oldArray` which are not included in `newArray` and not returned in the final result... :)

gtorodelvalle:
LGTM

Issue #349, opened 2016-05-19
If a device is registered using the library methods it must contain a service and a subservice


dmoranj:
Currently, the library checks the presence of the service and subservice headers for incoming requests in the provisioning API, but not for internal calls through the library methods. Checks should be applied for both systems.

PR #351, opened 2016-05-19
Task/single configuration mode

### LINKED ISSUE #17: Change the Identification of the Devices.

### LINKED ISSUE #260: Entity retrieval on incoming notifications or queries uses the entity name without the type

### LINKED ISSUE #263: Initial entity provisioning does not take configurations into account

### LINKED ISSUE #343: Change the way the Configurations work
In order to align with the old IoT Agents (maked as a need in production projects), the way Configurations  work must be changed. With the current behavior:
- For each subservice, multiple configurations can be defined. Each one is defined by its type.
- Whenever a device is provisioned, the device is associated to a configuration iff it belongs to the same subservice and has the same type.

With the new behavior, each Configuration will be associated to a single subservice, and the type will be marked optional both for the Configurations and the Devices.

This behavior may be reverted in the future, to reenable multiple configurations per service.

dmoranj:
Fixes #343
Fixes #263
Fixes #260
Fixes #17

This PR changes the way Devices are identified, by using both the Device ID and the (service, subserviceID) instead of supposing a globally unique DeviceID. This has impact in a huge part of the code.

This code comes with a new property in the configuration: `singleConfigurationMode`.


lib/services/devices/deviceRegistryMemory.js
```diff
@@ -35,26 +35,41 @@ var registeredDevices = {},
...
-    delete registeredDevices[id];
+function removeDevice(id, service, subservice, callback) {
+    var services = Object.keys(registeredDevices);
 
-    logger.debug(context, 'Removing device with id [%s].', id);
+    if (!subservice && !callback) {
+        callback = service;
+    }
+
+    for (var i = 0; i < services.length; i++) {
```
gtorodelvalle: This may cause unexpected behaviour in the case where the callback function is passed as the `service` parameter. The `length` property exists for functions and it indicates the number of arguments passed to the function. Consequently, it may only cause these unexpected behaviour in the case when a callback taking arguments is passed.

On the other hand, it seems that if no service is passed, the device is not deleted so maybe what is no longer needed is the code to control the arguments passed in case the `service` and `subservice` are required arguments.

lib/services/devices/deviceRegistryMemory.js
```diff
@@ -110,16 +128,20 @@ function listDevices(service, subservice, limit, offset, callback) {
...
 
-function getDevice(id, callback) {
-    if (registeredDevices[id]) {
-        callback(null, registeredDevices[id]);
+function getDevice(id, service, subservice, callback) {
+    if (!subservice && !callback) {
+        callback = service;
+    }
+
+    if (registeredDevices[service] && registeredDevices[service][id]) {
```
gtorodelvalle: I think this may lead to unexpected results in case the callback function is passed as the `service` argument :)

lib/services/devices/deviceService.js
```diff
@@ -205,45 +205,97 @@ function registerDevice(deviceObj, callback) {
...
 
-        if (!deviceData.type) {
-            deviceData.type = config.getConfig().defaultType;
+    function findConfigurationGroup(deviceObj, callback) {
+        function handlerGroupFind(error, group) {
+            if (error || !group) {
+                callback(null, null);
+            } else {
+                callback(null, group);
+            }
```
gtorodelvalle: The whole previous `if`clause seems to be "equivalent" to `callback(null, group)` (the only difference being that the second argument of the `callback` will be passed as `undefined` instead as `null`.

lib/services/devices/deviceService.js
```diff
@@ -292,13 +351,13 @@ function unregisterDevice(id, callback) {
...
 
-    config.getRegistry().get(id, function(error, device) {
+    config.getRegistry().get(id, service, subservice, function(error, device) {
         async.waterfall([
             async.apply(removeAllSubscriptions, device),
             processUnsubscribes,
             async.apply(registrationUtils.sendRegistrations, true, device),
             processContextUnregister,
-            async.apply(config.getRegistry().remove, id)
+            async.apply(config.getRegistry().remove, id, service, subservice)
```
gtorodelvalle: There are some occurrences of `async.apply` which could directly be substituted by `apply`.

test/unit/examples/deviceProvisioningRequests/provisionDuplicatedDev.json
```diff
@@ -0,0 +1,12 @@
...
+    {
+      "device_id": "Light1",
+      "protocol": "GENERIC_PROTO",
+      "service" : "smartGondor",
+      "service_path": "gardens",
+      "entity_name": "TheFirstLight",
+      "entity_type": "TheLightType"
+    }
+  ]
+}
```
gtorodelvalle: Final blank line appreciated :)

test/unit/examples/groupProvisioningRequests/provisionDuplicateGroup.json
```diff
@@ -0,0 +1,35 @@
...
+      "static_attributes": [
+        {
+          "name": "bootstrapServer",
+          "type": "Address",
+          "value": "127.0.0.1"
+        }
+      ]
+    }
+  ]
+}
```
gtorodelvalle: Final blank line appreciated :)

test/unit/examples/groupProvisioningRequests/provisionFullGroup.json
```diff
@@ -0,0 +1,36 @@
...
+      "static_attributes": [
+        {
+          "name": "bootstrapServer",
+          "type": "Address",
+          "value": "127.0.0.1"
+        }
+      ]
+    }
+  ]
+}
```
gtorodelvalle: Final blank line appreciated :)

lib/services/devices/deviceRegistryMemory.js
dmoranj: I don't think so. It seems you are mixing the `service` parameter with the `services` internal variable. Look at line 61: the `services` object is an array, coming from the execution of the `Object.keys()` function. 

Regarding the second objection, in the case no service is passed, the device is removed from all the services.

lib/services/devices/deviceRegistryMemory.js
dmoranj: You are right... That shouldn't be the case, actually. One of the overall features of this PR is to make the service and subservice used along the full stack, so there shouldn't be any calls without service. I have filed an issue to add checks for all the internal methods, [here](https://github.com/telefonicaid/iotagent-node-lib/issues/349). Requests from the Provisioning API are required to have the fiware-XXX headers.

All that said, I'll remove the guards against service and subservice in the registry methods.

lib/services/devices/deviceService.js
dmoranj: You're right. Fixed it.

lib/services/devices/deviceRegistryMemory.js
gtorodelvalle: Ups, my fault, you're completely right... :) PR should not be reviewed beyond 17:00... :D
NTC

README.md
```diff
@@ -14,54 +14,164 @@
...
+
+The following sequence diagram shows the different NGSI interactions an IoT Agent makes with the Context Broker,
+explained in the following subsections (using the example of a OMA Lightweight M2M device).
 
 ![General ](https://raw.github.com/dmoranj/iotagent-node-lib/develop/img/ngsiInteractions.png "NGSI Interactions")
 
-Be aware that the IoT Agents are only required to support NGSI10 operations `updateContext` and `queryContext` in their standard formats (both in XML and JSON) but will not answer to NGSI9 operations (or NGSI convenience operations of any kind).
+Be aware that the IoT Agents are only required to support NGSI10 operations `updateContext` and `queryContext` in their
+standard formats (currently in JSON format; XML deprecated) but will not answer to NGSI9 operations
+(or NGSI convenience operations of any kind).
```
XavierVal: This requirement only apply to IoTAgents, it does not apply to the way CBroker info is requested because CBroker will convert the query/update to the proper NGSI IoTagent needs, isn't it?

lib/services/devices/deviceService.js
dmoranj: Fixed

README.md
dmoranj: Yes, that's the idea. The Context Broker converts all NGSI operations to standard operations when redirecting to Context Providers, to prevent CPs having to implement the full NGSI API.

gtorodelvalle:
LGTM on my side... :)

XavierVal:
LGTM here too 🌴

PR #354, opened 2016-05-20
FIX Device searching across multiple services

### LINKED ISSUE #353: Allow the getDevicesByAttribute() function of the In-Memory repository to search through all the services
Current implementation of the getDevicesByAttribute() forces the `service` attribute to be non-null, or it will raise an error (that should be controlled, as pointed out in other issue). This behavior should be changed, to match the one in the MongoDB Registry: if a service is specified, the result should match the service; if none is specified, all the devices matching the condition should be returned (regardless of their service).

dmoranj:
Fixes #353


AlvaroVega:
LGTM

PR #356, opened 2016-05-20
FIX Pass service and subservice for handlers

### LINKED ISSUE #355: Pass service and subservice as arguments in command handlers
Current command handlers receive just the type and id of the entity as parameters, but it should also pass service and subservice to be able to retrieve information about the device.

dmoranj:
Fixes #355


gtorodelvalle:
LGTM

PR #358, opened 2016-05-24
Update documentation with mandatory service and subservice

### LINKED ISSUE #352: Update the documentation to reflect the changes in function signatures
With the last refactor, most of the IOTA function signatures where changed to add the service and subservice parameters to the call, but this change was not reflected in the documentation.

dmoranj:
Fixes #352

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,34 +298,34 @@ If the device has been previously preprovisioned, the missing data will be compl
 ##### iotagentLib.unregister()
 ###### Signature
 ```
-function unregisterDevice(id, callback)
+function unregisterDevice(id, service, subservice, callback)
 ```
 ###### Description
 Unregister a device from the Context broker and the internal registry.
 ###### Params
  * id: Device ID of the device to register.
+ * service: Service of the device to unregister.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -443,59 +449,72 @@ The handler is expected to call its callback once with no parameters (failing to
 function setConfigurationHandler(newHandler)
 ```
 ###### Description
-Sets the new user handler for the configuration updates. This handler will be called every time a new configuration is created or an old configuration is updated. 
+Sets the new user handler for the configuration updates. This handler will be called every time a new configuration is
+created or an old configuration is updated.
 
 The handler must adhere to the following signature:
 ```
 function(newConfiguration, callback)
```
gtorodelvalle: I guess the name for this function is missing... :)

dmoranj: Well, it's not missing, really. That's the prototype for the handler function the user have to pass to the `setConfigurationHandler()` function being described in that section. The user can use any name they want.

gtorodelvalle: NTC

gtorodelvalle:
LGTM

PR #359, opened 2016-05-24
Move defaultResource attribute to config root

dmoranj:
Move defaultResource attribute from the IoTManager section to config root (as it may be used in situations where an IoTManager is not needed).


XavierVal:
LGTM

PR #362, opened 2016-05-24
Task/add default api key utils

### LINKED ISSUE #360: Add defaultAPIKey and utils to retrieve the APIKey for a device
The IOTAs that use APIKeys for Southbound access may need to retrieve the effective APIKey associated to a device. Retrieving this information involves:

* searching for the group associated to the device if there is any
* returning a sensible default value if no group was found
* and returning an error in case none of the above can be achieved

This functionality is more or less implemented in the `getEffectiveApiKey()` function of the Ultralight IoT Agent, but it doesn't take into account:

* The possibility of working in a mode different than the singleConfigurationMode
* The possibility of working with types hardcoded in the configuration

The implementation can be move to the library, once those implementations have been addressed.

dmoranj:
Fixes #360

### DIFF
--- a/README.md
+++ b/README.md
@@ -837,7 +837,7 @@ have implications in the use of attributes with Context Providers, so this flag 
 * **singleConfigurationMode**: enables the Single Configuration mode for backwards compatibility (see description in the Overview). Default to false.
 * **timestamp**: if this flag is activated, the IoT Agent will add a 'TimeInstant' metadata attribute to all the attributes updateded from device information. 
 * **defaultResource**: default string to use as resource for the registration of new Configurations (if no resource is provided).
-
+* **defaultKey**: default string to use as APIKey for devices that do not belong to a particular Configuration.
 
 ## <a name="aboutapi"/> About API
 The library provides a simple operation to retrieve information about the library and the IoTA using it. A GET request
--- a/README.md
+++ b/README.md
@@ -541,6 +541,19 @@ Find a device group based on its service and subservice.
 * service: name of the service of the configuration.
 * subservice: name of the subservice of the configuration.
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -824,7 +837,7 @@ have implications in the use of attributes with Context Providers, so this flag
 * **singleConfigurationMode**: enables the Single Configuration mode for backwards compatibility (see description in the Overview). Default to false.
 * **timestamp**: if this flag is activated, the IoT Agent will add a 'TimeInstant' metadata attribute to all the attributes updateded from device information. 
 * **defaultResource**: default string to use as resource for the registration of new Configurations (if no resource is provided).
-
+* **defaultKey**: default string to use as APIKey for devices that do not belong to a particular Configuration.
```
gtorodelvalle: `APIKey` -> `API Key` :)

test/unit/provisioning/device-group-utils_test.js
```diff
@@ -0,0 +1,172 @@
...
+    };
+
+describe('Device Group utils', function() {
+    afterEach(function(done) {
+        iotAgentLib.deactivate(function() {
+            groupRegistryMemory.clear(done);
+        });
+    });
+
+    describe('When an APIKey is requested for a device in a group without the SingleConfiguration mode', function() {
```
gtorodelvalle: There are many appearances of `APIKey` all together in this file, I do not know if you want to keep it that way of changing it to `API Key`... :)

dmoranj: Fixed in 		a6e692e

README.md
dmoranj: Fixed in 		a6e692e

PR #364, opened 2016-05-25
FIX Mandatory update handler

### LINKED ISSUE #363: Command handler doesn't work if no update handler is defined
Current implementation fails when a updateContext requests comes from the Context Broker if there is a Command Handler but no Update Handler (for lazy attributes). An empty handler works to overcome this issue. This behavior should be fixed, so any of the handlers is optional.

dmoranj:
Fixes #363


AlvaroVega:
LGTM

PR #371, opened 2016-05-27
FIX Missing transport attribute in MongoDB registry

dmoranj:
FIX Missing transport attribute in MongoDB registry


manucarrace:
LGTM

PR #372, opened 2016-05-30
FIX Missing transport in MongoDB registers

dmoranj:
Missing transport attribute in MongoDB registers due to the order of insertion in the device provisioning handlers: modifying the value in the device object doesn't reflect in MongoDB (as the save instruction is executed before the handler).


XavierVal:
LGTM

Issue #397, opened 2016-06-16
Use the attributes defined in the Configuration as default alias values


dmoranj:
Mappings from object_id to name in configurations should be used as default values in the alias plugins, whenever there are no correspondent mappings provisioned in the device.

PR #398, opened 2016-06-20
Task/merge group configuration

dmoranj:
Implements #397

This PR changes the complete behavior of Configuration default values from "applied to the device provisioning and stored in the DB with it" to "calculated dynamically when needed and overwritten in the requests".


test/unit/lazyAndCommands/lazy-devices-test.js
```diff
@@ -169,7 +169,9 @@ describe('IoT Agent Lazy Devices', function() {
             async.series([
                 apply(iotAgentLib.activate, iotAgentConfig),
                 apply(iotAgentLib.register, device1)
-            ], done);
+            ], function(error) {
```
gtorodelvalle: Although it is a test, maybe you would like to leave it as it was before, I mean passing `done` as the callback to `async.series` :)

gtorodelvalle:
LGTM

PR #425, opened 2016-07-21
Bug/non ngsi response to ngsi requests

### LINKED ISSUE #422: IoTAgent returns non-NGSI error as a response to an error in the Context Providers
When a error is produced in the Context Provider operations, the IoT Agent returns an IoTAgent-specific type of error, that is not the expected response to an NGSI message.

dmoranj:
Fixes #422


AlvaroVega:
LGTM

PR #431, opened 2016-08-10
Task/iotam registration with static

### LINKED ISSUE #430: IOTAM Registrations won't send static attributes in group information
Current registrations for the IoTManager send only information about active attributes, missing both commands and static ones. This behavior should be fixed to make the registrations reflect all the information in the device groups.

dmoranj:
Fixes #430


AlvaroVega:
LGTM

PR #432, opened 2016-08-11
FIX Commands not supported by current IoTAM

dmoranj:
Remove Command support for IOTAM registration (not currently supported in C++ IoTAM).


XavierVal:
LGTM

PR #443, opened 2016-09-07
FIX Logger module not common with IoTAgents

### LINKED ISSUE #442: Logger modules not being singleton cause logging inconsistencies
Logops library is used both in the IoTAgent library, and in the IoTAgent code, but, by some mysterious reason, sometimes the modules are treated as a singleton (and changing the log level in one of them changes both) and sometimes they aren't and it just changes one of them. This should be changed, to avoid problems  in the IoTA Logs.

dmoranj:
Fixes #442


AlvaroVega:
LGTM

PR #446, opened 2016-09-09
FIX Remove Device ID index

### LINKED ISSUE #445: Remove Device ID Index
Up until now, the IoTAgent library supposed that the DeviceID was going to be unique globally accross the system. In a multitenant environment, though, that claim doesn't hold anymore, so the Unique index in the ID field of the Device model should be removed.

dmoranj:
Fixes #445


manucarrace:
LGTM

Issue #455, opened 2016-09-19
Add polling command support for the Library


dmoranj:
The library does not offer any mechanism for the implementation of Polling Commands, i.e.: commands that are not inmediatly sent to the device, but kept in a queue in the IoTAgent for its later execution. This mechanism has been requested by some customers, and should be implemented. 

The library will offer the following general functionality for the polling mechanism:
- Data repository for the stored commands.
- Functions for the management of stored commands.
- Automatic storage of incoming commands for devices marked as polling devices.

The library will not deal with any mechanisms regarding the South Bound APIs connecting with the command queue. Each independent IoTAgent will be in charge of setting devices as `polling devices` as support for this feature may vary from Agent to Agent.

PR #456, opened 2016-09-19
Task/polling command

dmoranj:
Implements #455


gtorodelvalle:
LGTM (probably not the best time of the day to do this though... :p)

dmoranj:
You can rethink your support tomorrow, we are not merging this request today.

lib/services/groups/groupRegistryMongoDB.js
```diff
@@ -260,7 +260,7 @@ function init(newConfig, callback) {
 }
 
 function clear(callback) {
-    dbService.dropDatabase(callback);
+    dbService.db.db.dropDatabase(callback);
```
AlvaroVega: is ```db``` duplicated ?

dmoranj: Yes, it is, due to a horrible choice of attribute names (by me). The first db is the attribute of the module containing the connection; the second one is the `db` attribute of the `connection` object, that holds a pointer to the actual database (as can be used in Mongo's command line interpreter).

AlvaroVega:
LGTM

PR #464, opened 2016-10-04
Task/startup retry

### LINKED ISSUE #357: when running service retry when checking for dependencies (mongo, orion, etc.)
In docker-compose scenarios or other situations it can be helpful to implement a retry mechanism (similar to what is done by Orion) so that if mongo or orion are not ready by the time the IoTA is running, the IoTA does not die immediately but retries for a limited number of times or period until the needed dependencies are running. 

This could happen because even docker-compose would run dependencies before, each component may have different start-up times and as a result race conditions could arise.

dmoranj:
Fixes #357


dmoranj:
Adds two new configuration parameters: mongodb.retries and mongodb.retryTime

PR #482, opened 2016-10-19
Task/bidirectionality for configurations

### LINKED ISSUE #479: Bidirectional plugin doesn't apply to active attributes defined in a Configuration Group
Definition of bidirectional values should be allowed for both devices and Configuration Groups. Current behavior of the plugin is to provision the appropriate subscriptions just in the case when the reverse expressions are declared directly for a device.

dmoranj:
Fixes #479


AlvaroVega:
LGTM

PR #511, opened 2016-11-03
Bug/multiple variables with numbers

### LINKED ISSUE #509: Pool command expired but CB not updated
When a poll command is expired (the device has not requested the command within the valid period) the CB is not updated accordingly. Iotagent should update CB with command_status ERROR and command_info with a descriptive error message.

dmoranj:
Fixes #509


XavierVal:
LGTM

PR #513, opened 2016-11-07
Task/alarm management sysyem

### LINKED ISSUE #512: Add an alarm management module
Add a module to hold information of the alarms that have been thrown during the execution of the component. This module should provide:

- A mechanism to raise an alarm. The alarm should be identified by a string code. E.g.: MONGODB_ALARM_01.

- A mechanism to release an alarm.

Raising and releasing an alarm should generate an ERROR and INFO log respectively.

It should also provide a function decorator that can be used to intercept calls to modules whose execution without error should be considered an alarm releasing trigger. E.g.: any correct execution of a method in the DeviceRegistryMongoDB module should release any MONGODB_ALARM_01 alarms.

dmoranj:
Fixes #512


AlvaroVega:
LGTM

PR #515, opened 2016-11-08
ADD Shared constants and errors

### LINKED ISSUE #386: Add a mechanism to share constants and error definitions between IoTAs
There are some constants and some error definitions that are commonly used along all the IoTAs (and some of them shared with this library), and are currently replicated across all the IoTA constant and error files. Considering all the IoTA and error files have the same structure, a mechanism could be proposed to  inherit the Lib constants and errors, so that there is no need to replicate that information among the different IoTAs.

dmoranj:
Fixes #386


AlvaroVega:
LGTM

XavierVal:
LGTM

PR #522, opened 2016-11-16
FIX alarm raised when callback called with null

### LINKED ISSUE #521: Alarm raises with null error
When a function intercepted by the alarm management interceptor calls its callback with a null error, the interceptor raises an alarm (without need).

dmoranj:
Fixes #521


AlvaroVega:
LGTM

PR #524, opened 2016-11-17
Capture MongoDB connection events

dmoranj:
Non-captured MongoDB connection errors made the IoTA stall when the MongoDB server went down. Debug logs have also been added to other events to monitor changes in the connection status.


AlvaroVega:
LGTM

PR #531, opened 2016-12-23
Updated iotAgentLib.getDevice call

vertigo7x:
Hi!
In the call to iotAgentLib.getDevice some parameters are missing.
Set them to null.
Thanks for your great work!


dmoranj:
Thanks for your contribution.

Issue #532, opened 2017-01-03
Unify alarm release and raise log levels


dmoranj:
With the current version of the library, whenever an alarm is raised, a ERROR log is logged to the output, while the release of the alarm causes an INFO level log. Both should be unified either to ERROR or to LOG levels.

PR #537, opened 2017-01-19
log releasing alarm as error instead of info in order to not loose it…

AlvaroVega:
log releasing alarm as error instead of info in order to not loose it when loglevel is error

Fixes: https://github.com/telefonicaid/iotagent-node-lib/issues/532


rfgtid:
LGTM

Issue #527, opened 2016-12-05
Support for NGSIv2


dmoranj:
Currently, all the IoTAgents communicate with the Context Broker using NGSIv1 for all the communications (both for context provisioning and context management). A migration to NGSIv2 is needed, as some behaviors are starting to diverge (mainly the use of special types in NGSIv2).

dcalvoalonso:
We propose to create a first functional version to support the [Release Candidate version of NGSIv2](http://telefonicaid.github.io/fiware-orion/api/v2/stable/). This first version should be a Minimum Viable Product (MVP) that provides a relevant set of NGSIv2 functionalities.

The following modifications should be added:
* Communication with Context Broker considers new URLs and payloads (following RESTful principles).
  * Entity Creation.
  * Query entity. Including options to retrieve keyValues, values for a list of attributes or a single attribute.
  * Text-based attribute value set/get
  * Getting all entities and filtering.
  * Update entity attributes and a single attribute.
  * Creation of subscriptions.
  * Pause of subscriptions.
  * Delete subscriptions.
* Native JSON datatypes.
* Datetime support based on ISO ISO8601 format.
* Communication with Context Broker considers new error responses.
* Unit tests and examples cover completely the new modifications (100% of code coverage).

One important question to be answered is if we should maintain compatibility with NGSIv1 in the IoT Agents.  If we do so, we should introduce an option in the configuration to choose which version of NGSI to use. Obviously, the complexity of the resulting code will be higher.

The next step would be the migration of [iot.IoTagent-UL](https://github.com/Fiware/iot.IoTagent-UL), [iot.IoTagent-JSON](https://github.com/Fiware/iot.IoTagent-JSON) and [iot.IoTagent-LWM2M](https://github.com/Fiware/iot.IoTagent-LWM2M).

fgalan:
@dcalvoalonso thank you for your contribution to issue comments!

Please, take into account the following feedback:

* With regards to NGSIv2 operations, note that IOTAs use only a subset of the API. In particular, create/pause/delete subscriptions and get all entities and filtering are not used by ITOAs. I have also doubts with regards to text-based set/get... probablly it is also unneeded. The basic NGSI operations IOTA uses are:
[...]

PR #555, opened 2017-10-23
Support for NGSIv2

dcalvoalonso:
Fixes: [527](https://github.com/telefonicaid/iotagent-node-lib/issues/527)

First version that supports NGSIv2. 
Some aspects are not covered yet:
* Registrations to CB : specified in the last version of NGSI but not implemented in CB)
* Acts as Context Provider for queries/updates coming from CB (for passive attributes and commands): Not specified yet


lib/fiware-iotagent-lib.js
```diff
@@ -177,20 +199,115 @@ function deactivate(callback) {
...
+exports.unregister = unregister;
+exports.query = query;
+exports.update = update;
+exports.setCommandResult = setCommandResult;
+exports.listDevices = listDevices;
+exports.getDevice = getDevice;
+exports.getDeviceByName = getDeviceByName;
+exports.getDevicesByAttribute = getDevicesByAttribute;
+exports.mergeDeviceWithConfiguration = mergeDeviceWithConfiguration;
+exports.retrieveDevice = retrieveDevice;
```
jmcanterafonseca: are these changes really needed?

lib/fiware-iotagent-lib.js
```diff
@@ -177,20 +199,115 @@ function deactivate(callback) {
     contextServer.stop(callback);
 }
 
+function update(entityName, attributes, typeInformation, token, callback) {
+    ngsi.update(entityName, attributes, typeInformation, token, callback);
+}
+
+function query(entityName, attributes, typeInformation, token, callback) {
+    ngsi.query(entityName, attributes, typeInformation, token, callback);
+
```
jmcanterafonseca: not needed line, please use a Linter

lib/fiware-iotagent-lib.js
```diff
@@ -177,20 +199,115 @@ function deactivate(callback) {
...
+function setCommandResult(entityName, resource, apikey, commandName,
+                          commandResult, status, deviceInformation, callback) {
+    ngsi.setCommandResult(entityName, resource, apikey, commandName,
+                          commandResult, status, deviceInformation, callback);
+
+}
+
+function addUpdateMiddleware(middleware) {
+    ngsi.addUpdateMiddleware(middleware);
+
```
jmcanterafonseca: not needed line

lib/fiware-iotagent-lib.js
```diff
@@ -199,23 +316,23 @@ exports.setDataQueryHandler = contextServer.setQueryHandler;
 exports.setConfigurationHandler = contextServer.setConfigurationHandler;
 exports.setProvisioningHandler = contextServer.setProvisioningHandler;
 exports.setNotificationHandler = contextServer.setNotificationHandler;
-exports.addUpdateMiddleware = ngsi.addUpdateMiddleware;
-exports.addQueryMiddleware = ngsi.addQueryMiddleware;
+exports.addUpdateMiddleware = addUpdateMiddleware;
```
jmcanterafonseca: are these changes really needed?

lib/fiware-iotagent-lib.js
```diff
@@ -235,6 +352,15 @@ exports.dataPlugins = {
     bidirectionalData: require('./plugins/bidirectionalData')
 };
 
+exports.dataPluginsNgsi2 = {
+    attributeAlias: require('./pluginsNgsi2/attributeAlias'),
```
jmcanterafonseca: for NGSIv2 you don't need to create a new plugin

lib/pluginsNgsi2/addEvent.js
```diff
@@ -0,0 +1,37 @@
...
+ * See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with fiware-iotagent-lib.
+ * If not, seehttp://www.gnu.org/licenses/.
+ *
+ * For those usages not covered by the GNU Affero General Public License
+ * please contact with::daniel.moranjimenez@telefonica.com
+ *
+ * Modified work Copyright 2017 Atos Spain S.A
```
jmcanterafonseca: this is not the way to state your authorship. You can add a sentence like

`Author: Daniel Calvo Atos Spain S.A. ` 

IPR issues are currently being discussed by Fiware Foundation and current IPR owners

lib/pluginsNgsi2/attributeAlias.js
```diff
@@ -0,0 +1,111 @@
+/*
```
jmcanterafonseca: do we really need this new component?

lib/pluginsNgsi2/bidirectionalData.js
```diff
@@ -0,0 +1,276 @@
+/*
```
jmcanterafonseca: do we really need this new component?

lib/fiware-iotagent-lib.js
mrutid: Indeed, not needed code... why this? (and all the rest, seems pretty much out of scope of the PR)

lib/pluginsNgsi2/addEvent.js
fgalan: An example (currently used at Orion repository) of the style proposed by @jmcanterafonseca:

https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/serviceRoutinesV2/optionsGetOnly.cpp#L23

lib/pluginsNgsi2/expressionParser.js
```diff
@@ -0,0 +1,202 @@
+/*
```
jmcanterafonseca: is this new code? or just moving an existing component?

lib/pluginsNgsi2/expressionPlugin.js
```diff
@@ -0,0 +1,86 @@
+
```
jmcanterafonseca: is the expression plugin different than the NGSIv1 one?

lib/pluginsNgsi2/expressionPlugin.js
```diff
@@ -0,0 +1,86 @@
...
+
+function update(entity, typeInformation, callback) {
+    try {
+        var entityAtts = [];
+        var expressionAttributes = [];
+        entityAtts = utils.createArrayFromAttributes(entity);
+        var context = parser.extractContext(entityAtts);
+        if (typeInformation.active) {
+            expressionAttributes = parser.processExpressionAttributes(typeInformation, typeInformation.active, context);
```
jmcanterafonseca: line seems to be very long. check with a linter

lib/pluginsNgsi2/pluginUtils.js
```diff
@@ -0,0 +1,171 @@
...
+ * For those usages not covered by the GNU Affero General Public License
+ * please contact with::daniel.moranjimenez@telefonica.com
+ *
+ * Modified work Copyright 2017 Atos Spain S.A
+ */
+
+'use strict';
+
+function createProcessAttribute(fn, attributeType) {
+    return function(attribute) {
```
jmcanterafonseca: new line not needed

lib/fiware-iotagent-lib.js
```diff
@@ -116,6 +130,14 @@ function doActivate(newConfig, callback) {
         commandRegistry = require('./services/commands/commandRegistryMemory');
     }
 
+    if (checkNgsi2(newConfig)) {
+        logger.info(context, 'Using NGSIv2');
+        ngsi = ngsi2;
+        deviceService = deviceServiceNgsi2;
+        commands = commandsNgsi2;
+        subscriptions = subscriptionsNgsi2;
```
fgalan: Note IOTAs don't uses any subscription... this seems to be over-implemented. The three NGSIv2 interactions that should be supported are only the ones described in #527 

1. Registration to CB on device provision time.
2. Send updates to CB (for active attributes).
3. Acts as Context Provider for queries/updates coming from CB (for passive attributes and commands).

Of those, only point 2 is achiavable now (as 1 and 3 are not implemented/specified yet). I'd suggest to reduce the scope just to that point. That should reduce the size of the PR and make it reviewable.

Not sure of the NGSIv1-based code is structured, but if the POST /v1/updateContext is modularized in just one point (where the verb + url + payload is composed) maybe is a matter of just adapting that part, creating a NGSIv2 version for that and switching to the new/old method depending on the parameter in config.js.

lib/services/common/domain.js
```diff
@@ -54,7 +54,7 @@ function cleanDomain(domainToClean) {
 
 function generateDomainErrorHandler(domainToHandle) {
     return function domainErrorHandler(err) {
-        logger.error(context, err);
+        logger.error('aaaaaa'+context, err);
```
mrutid: Fix log message

lib/services/common/genericMiddleware.js
```diff
@@ -108,6 +108,18 @@ function ensureType(req, res, next) {
 }
 
 /**
+ * Ensures the request type is one of the supported ones.
+ */
+function ensureTypeNgsi2(req, res, next) {
```
mrutid: It's ensuring content type==JSON, not Ngsiv2

lib/services/northBound/restUtils.js
```diff
@@ -148,8 +148,19 @@ function isTimestamped(payload) {
     return false;
 }
 
+function isTimestampedNgsi2(payload) {
```
mrutid: Please check "payload.hasOwnProperty(i)" when iterating over objects

lib/pluginsNgsi2/addEvent.js
fgalan: In the case of an existing file, "Modified by:" would be better.

dcalvoalonso:
Thanks for your prompt review and comments, I really appreciate the time and effort you have spent to give me feedback :).
I will try to clarify some points below:
* I will proceed as you propose and I will split the changes in several PRs to simplify the review process.
* I will add an ownership sentence as you propose instead of the Modified work Copyright statement:  Modified by:: Daniel Calvo - ATOS Research & Innovation 
* The approach I have followed is to try to maintain as clean as possible the original NGSIv1 code. For instance, if we talk about how to send updates for active attributes to the CB, I see that we have two options:
  1. To add a new parameter to sendUpdateValue function in ngsiService.js to choose between NGSIv1 and NGSIv2. This would imply modifying also other functions like generateNGSIOperationHandler. From my point of view, the resulting code would be more complex but of course, it is an option.
  2. To create a new version of ngsiService that implements specific code for NGSIv2. In fiware-iotagent-lib, as part of the activation and configuration loading, we check if NGSIv2 is used and use the corresponding ngsiService. In my eyes, this solution is less intrusive in the original code and the migration and support for new NGSIv2 related functionalities would be simpler. I have followed this alternative.
* Regarding plugins, I guess that the discussion is more or less similar to the previous point. Do we need a new plugin for NGSIv2? Well, we can add a new parameter in some functions to decide if we have to use NGSIv1 or NGSIv2 (for instance, updatePlugin function of timestampProcessPlugin.js) or just isolate the new specific code for NGSIv2 in a separate plugin. Both options are feasible and I am open to your proposals.
* Regarding subscriptions in IOTAs. The current version does implement subscriptions using NGSIv2:
  1. https://github.com/Fiware/iot.IoTagent-node-lib/blob/master/lib/services/ngsi/subscriptionService.js
  2. https://github.com/Fiware/iot.IoTagent-node-lib/blob/master/test/unit/ngsiService/subscriptions-test.js
  3. As you can see, I understand from the code that the idea is that a client using an IOTA can send a subscription request to the CB from the IOTA.
 * In general, what I have tried is to replicate all the functionalities that are implemented in the current version for NGSIv1 for NGSIv2.
* Regarding passing Travis-CI before merging... I completely agree with you... but if you check the job log you can see that tests failing are for NGSIv1. If you directly clone the master branch of the repository and run npm install and npm test, the output is the following:
```
263 passing (27s)
  4 pending
  20 failing
```

fgalan:
With regards to approach and plugins, from my point of view having two plugins (one for NGSIv1 and another NGSIv2) or any other mechanism of "parallel" code could make senses in the case we want NGSIv1 and NGSIv2 at the end. However, note that NGSIv2 and NGSIv1 funcionality will coexits for short time (as we [dicussed in the root issue](https://github.com/telefonicaid/iotagent-node-lib/issues/527#issuecomment-337565075)). At the end, only NGSIv2 will remain. In this sense, I think that the complexity introduced in sendUpdateValue and other functions will be temporal, thus makes sense going on that line (as I understand new PR #556 does).

With regards to subscription, you are rigth and I was wrong :) I don't know why that code is there (maybe it comes from the old days, when it was not so clear how to integrate IOTA and CB) but it is not used. I'd create and issue about removing that code. There is not need of migrating that operation to NGSIv2.

With regards to travis, currently master is ok at https://travis-ci.org/telefonicaid/iotagent-node-lib/branches. Maybe the problems is the one that @AlvaroVega point out at https://github.com/telefonicaid/iotagent-node-lib/pull/556/#discussion_r147087022

dcalvoalonso:
@fgalan, I will proceed then as you suggest for plugins, I think that it's a good approach! :)

Regarding travis, I have seen that @AlvaroVega has the same problem in #557  . As you can see, the unit tests failing are exactly the same. :(

fgalan:
> With regards to subscription, you are rigth and I was wrong :) I don't know why that code is there (maybe it comes from the old days, when it was not so clear how to integrate IOTA and CB) but it is not used. I'd create and issue about removing that code. There is not need of migrating that operation to NGSIv2.

Issue created https://github.com/telefonicaid/iotagent-node-lib/issues/558

fgalan:
Finally, the work was done in PR #556 . I understand this PR can be closed. Is that correct, @dcalvoalonso ?

dcalvoalonso:
Yes, I agree with you! :+1:

PR #591, opened 2018-03-26
Add oauth 2.0 support for authentication against context broker

chicco785:
* Add oauth 2.0 support for authentication agains context broker
* Introduce support for https in authentication endpoint using `url` configuration parameter
* Allow dynamic configuration of authentication header


config.js
```diff
@@ -32,6 +32,9 @@ var config = {
         host: '0.0.0.0'
     },
     authentication: {
+        enabled: true,
```
fgalan: Two questions with regards the new parameters:

* There should be an environment variable corresponding to each one of them (properly processed in the code and documented). Have they been included?
* Thinkinig in backward compatilbity, which are the default vlues for the 3 new elements in the case they doesn't appear in config.js/envars?

doc/installationguide.md
```diff
@@ -21,21 +21,51 @@ These are the parameters that can be configured in the global section:
...
-        host: 'localhost',
-        port: '5000',
-        user: 'iotagent',
-        password: 'iotagent'
-	}
-```
+* **authentication**: authentication data, for use in retrieving tokens for devices with a trust token (required in scenarios with security enabled in the Context Broker side). 
+  Currently, two authentication provider are supported: `keystone` and `oauth2`. 
+  In `keystone` based authentication, the `trust` associated to the `device` or `deviceGroup` is a token representing a specific user and his rights on a given domain (i.e.
+  combination of `fiware-service` and `service-path`). The authentication process use the trust delegation workflow to check if the trust provided is valid, in which
```
fgalan: I understant that `service-path` refers to the name of the header. In that case, the right name is `fiware-servicepath`

config.js
chicco785: hi @fgalan auth parameters where not handled as env variables before, so i didn't work that out.

regarding backward compatibility, given that url when missing is computed form host and port that would not be an issue. also, if enabled and type not specified, assumption is that it is 'keystone' the provider used. PS: enabled was already there, but not documented.

doc/installationguide.md
chicco785: correct :) btw I hope my understanding of the usage of domain for auth in keystone is correct. will fix it.

doc/installationguide.md
```diff
@@ -21,21 +21,51 @@ These are the parameters that can be configured in the global section:
...
-* **authentication**: authentication data, for use in retrieving tokens for devices with a trust token (just needed in scenarios with security enabled in the Context Broker side). E.g.:
-```
-	{
-        host: 'localhost',
-        port: '5000',
-        user: 'iotagent',
-        password: 'iotagent'
-	}
-```
+* **authentication**: authentication data, for use in retrieving tokens for devices with a trust token (required in scenarios with security enabled in the Context Broker side).
```
fgalan: `enabled` is not include in this documentation

doc/installationguide.md
```diff
@@ -21,21 +21,51 @@ These are the parameters that can be configured in the global section:
...
+    ```
+    {
+          enabled: true,
+          url: 'https://localhost:5000',
+          type: 'keystone',
+          user: 'iotagent',
+          password: 'iotagent'
+    }
+    ```
+  In `oauth2` based authentication, the `trust` associated to the `device` or `deviceGroup` is a `refresh_token` issued by a specific user for the context broker client.
```
fgalan: context broker -> Context Broker

(to use the same style used above)

config.js
chicco785: ps: also header default value is the old one and is loaded from a new constants

chicco785: as regards pre-existence of `enabled` see: https://github.com/telefonicaid/iotagent-node-lib/pull/591/files#diff-31cabb4f8b54ad34fdf6b4b99d056f49L346

doc/installationguide.md
fgalan: (In general apply the same to other occurrences of the same in the document)

doc/installationguide.md
```diff
@@ -21,21 +21,51 @@ These are the parameters that can be configured in the global section:
...
+          user: 'iotagent',
+          password: 'iotagent'
+    }
+    ```
+  In `oauth2` based authentication, the `trust` associated to the `device` or `deviceGroup` is a `refresh_token` issued by a specific user for the context broker client. 
+  The authentication process use the [`refresh_token` grant type](https://tools.ietf.org/html/rfc6749#section-1.5) to obtain an `access_token`
+  that can be used to authenticate the request to the context broker.
+  At the time being the assumption is that the `refresh_token` is a not expiring `offline_token` (we believe this is the best solution in the case of IoT Devices,
+  since injecting a refresh token look may slow down communication. Still the developer would be able to invalidate the refresh token on the provider side in
+  case of security issues connected to a token). The code was tested using Keycloak and Auth0 (it may require customisation
```
fgalan: It would be useful to include link to Keycloak and Auth0 (using the `[...](...)` syntax).

chicco785: fix by 9ea2773

doc/installationguide.md
chicco785: fix by 9ea2773

doc/installationguide.md
chicco785: fix by 9ea2773

doc/installationguide.md
chicco785: fix by 9ea2773

test/unit/general/contextBrokerKeystoneSecurityAccess-test.js
```diff
@@ -38,6 +38,7 @@ var iotAgentLib = require('../../../lib/fiware-iotagent-lib'),
             port: 4041
         },
         authentication: {
+            type: 'keystone',
```
chicco785: @fgalan to proof backward compatibility, i can remove this line ;)

lib/services/ngsi/ngsiService.js
```diff
@@ -318,6 +318,22 @@ function sendQueryValue(entityName, attributes, typeInformation, token, callback
...
+ *
+ * @param {Object} deviceGroup        The deviceGroup where the trust token was stored. Null if stored in actual device.
+ * @param {Object} deviceInformation  The device where the trust token was stored. Null if stored in deviceGroup.
+ * @param {String} trust              The current trust token.
+ * @param {Object} response           The response where the new token is stored.
+ * @param {Function} callback         The callback function.
+ */
+
+function updateTrust(deviceGroup, deviceInformation, trust, response, callback) {
+    callback(null, response);
```
chicco785: @fgalan this is a place holder, in case we have an authentication mechanism that require update of `trust`. can be removed if you think it is not worth to keep it at the time being.

config.js
chicco785: 1. let me know if current backward compatibility is fine or any changes are needed
1. let me know if despite before env variables where not used for authentication you want me to introduce them.

doc/installationguide.md
```diff
@@ -21,21 +21,51 @@ These are the parameters that can be configured in the global section:
...
+  for other providers - while OAuth2 is a standard, not all implementations behave in the same way, especially as regards status codes and error messages).
+  Required parameters are: the `url` of the OAuth 2 provider to be used (alternatively `host` and `port` but if you use this combination, the IoT Agent will assume
+  that the protocol is HTTP), the `tokenPath` to which the validation request should be sent, the `clientId` and `clientSecret` that identify the context broker,
```
fgalan: Reading

> the `clientId` and `clientSecret` that identify the context broker

it seems that Context Broker acts as client in the OAuth athentication. However, I'd say that IOTA takes that role (because is the one that sends the authentication request). Should the documentation be improved? Or maybe I'm missunderstanding how it works (I'm not an expert in OAuth)

lib/command/commandLine.js
```diff
@@ -887,6 +888,8 @@ var commands = {
 function init(cbCfg, iotCfg) {
     config = cbCfg;
     configIot = iotCfg;
+    if(!config.authorization && !config.authorization.header)
```
fgalan: The style we typically use is:

```
if (...) {
  ...
}
```

I suggest using the same here.

doc/installationguide.md
chicco785: hi @fgalan OAuth authentication supports different type of flows that allows for different type of authentication.
in this case, being the purpose to have a token specific for the service (each service / application supported by an OAuth provider can be issued specific tokens), we use the [client credentials flow](https://tools.ietf.org/html/rfc6749#section-1.3.4). In this picture, the client (IOTA) authenticate to the OAuth provider using a `clientId` and `clientSecret` that are specific to the service to be accessed (i.e. the context broker). Naming are the conventional names adopted in OAuth.

lib/constants.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     TIMESTAMP_TYPE: 'ISO8601',
     SERVICE_HEADER: 'fiware-service',
     SUBSERVICE_HEADER: 'fiware-servicepath',
+    AUTH_HEADER: 'X-Auth-Token',
```
fgalan: Maybe it should be used in lowercase, as the other headers?

chicco785: mmm... not sure, will keystone complain?

lib/fiware-iotagent-lib.js
```diff
@@ -116,6 +129,19 @@ function doActivate(newConfig, callback) {
...
     logger.info(context, 'Activating IOT Agent NGSI Library.');
 
+    if (newConfig.authentication && newConfig.authentication.enabled) {
+        if (newConfig.authentication.type &&
+            newConfig.authentication.type === 'oauth2') {
+            logger.info(context, 'OAuth2 selected as authentication service');
+
+            securityService = require('./services/common/securityServiceOAuth2');
+        } else {
+            logger.info(context, 'Falling back to Keystone as authentication service');
```
fgalan: I agree that if authentication.type is not provided is a "Failling back" case, but if the user explicitely configured authentication.type === 'keystone' is a "Keystone selected case".

This problem could be avoided with a slight change in the log messages:

"Using OAuth2 as authentication service"
"Using Keystone as authentication service"

chicco785:
@fgalan i just realised that for subscriptions created by the IOTA, authorisation is not enabled... am I wrong? in case this would land in a different PR

lib/command/commandLine.js
chicco785: fixed by 68f37d7

lib/fiware-iotagent-lib.js
chicco785: fixed by b2682c7

test/unit/general/contextBrokerKeystoneSecurityAccess-test.js
fgalan: Good idea. But instead of completealy remove it, just remove it, i.e:

```
// Next line is syntactically correct from a configuration point of view, but we comment it in order to ensure backward compability
// type: keystone
```

or a similar wording.

lib/constants.js
fgalan: Goot point. I'm almost sure it will work, but need to be checked. However, let's do it in a separete PR.

In the contet of this PR, a `//FIXME: check Keystone support this in lowercase, then change` would be enough.

chicco785: fix by 871e6ef

test/unit/general/contextBrokerKeystoneSecurityAccess-test.js
chicco785: fix by 871e6ef

chicco785:
this branch is now rebased and linted (except for some camel case errors, that I believe I cannot change due to the expected field names in oauth body request)

chicco785:
58ede8b fix the lint error for camelcase by adding a notation before the functions that generated the error

config.js
fgalan: 1. According to your information, backward compatibility seems fine.
2. No need to add env var (at least in the context of this PR) if they have been never used for this kind of configuration.

NTC

lib/services/ngsi/ngsiService.js
fgalan: I think is a good idea.

NTC

doc/installationguide.md
fgalan: I'm not an expert in OAuth ;) but if people familiar with find that nomenclature is ok, then fine with it.

NTC

lib/fiware-iotagent-lib.js
```diff
@@ -116,6 +130,19 @@ function doActivate(newConfig, callback) {
 
     logger.info(context, 'Activating IOT Agent NGSI Library.');
 
+    if (newConfig.authentication && newConfig.authentication.enabled) {
+        if (newConfig.authentication.type &&
+            newConfig.authentication.type === 'oauth2') {
+            logger.info(context, 'Using OAuth2 as authentication service');
+
+            securityService = require('./services/common/securityServiceOAuth2');
```
fgalan: We typically set require() at the beginning of the file. Thus, instead of ursing require() in this point (and L140) use the corresponding var.

Except if there is a strong reason to have the require defined in the middle...

lib/services/common/securityServiceKeystone.js
```diff
@@ -89,4 +89,26 @@ function getToken(trust, callback) {
...
+function getToken(trust, response, callback) {
+
+    logger.debug(context, 'Retrieving token from the KEYSTONE response');
+
+    if (response.headers['x-subject-token']) {
+        logger.debug(context, 'Token found [%s] for trust [%s]', response.headers['x-subject-token'], trust);
+
+        callback(null, response.headers['x-subject-token']);
+    } else {
+        logger.error(context, 'KEYSTONE-003: Token missing in the response headers', response.headers);
```
fgalan: KEYSTONE-003 is a new error message? Or you are reusing an existing one?

In the case of being a new one, I think we have a place in the documentation where it should be added.

test/unit/general/contextBrokerOAuthSecurityAccess-test.js
```diff
@@ -0,0 +1,256 @@
+/*
```
fgalan: I haven't review this in detail, as I'm not an Oauth expert... but maybe somebody should do...

NTC (just informative)

fgalan:
I have done a new review. Excetp for a couple of minor things (see my 2 last in-code comments) I think is almost done. Good! :)

@dcalvoalonso what do you think? Looks good for you?

lib/services/common/securityServiceKeystone.js
chicco785: new one, i will update doc

lib/fiware-iotagent-lib.js
chicco785: i think i did in the same way it was done for dynamic linking of registries. let me check

chicco785: see: https://github.com/telefonicaid/iotagent-node-lib/blob/e542d79f94ad2a5b129bd817b85d9b7c30ddbb39/lib/fiware-iotagent-lib.js#L124

dcalvoalonso:
In my eyes it is OK. When your final comments are solved and the conflict regarding the CNR fixed, everything should be ready.

lib/services/common/securityServiceKeystone.js
chicco785: fixed by 57ede1a

chicco785:
conflict fixed cf2ad00f47b8673220ea2107ae337a47cf9c7c7e

PR #597, opened 2018-03-26
remove devices related to a service when device=true option used (see #596)

### LINKED ISSUE #596: missing support for device=true in the delete service endpoint
According to the documentation of the Configuration API, when removing a service the following parameter should allow to remove all attached devices to the service:
* `device (optional, boolean)`. Default value is false. Remove devices in service/subservice. This parameter is not valid when Fiware-ServicePath is '/*' or '/#'.

chicco785:
@fgalan @dcalvoalonso 
this is an attempt to fix #596, but i need a some support in understanding why i get an ECONNECT error from the nock server when unregistration are sent to context broker.

i supposed that:

```
            contextBrokerMock = nock('http://192.168.1.1:1026')
                .matchHeader('fiware-service', 'TestService')
                .matchHeader('fiware-servicepath', '/testingPath')
                .post('/NGSI9/registerContext')
                .reply(200,
                utils.readExampleFile(
                   './test/unit/examples/contextAvailabilityResponses/unregisterDevice1Success.json'));
```

would have been the only thing needed, but apparently something is still wrong (when using a real context broker there is no issue).

thx.

### DIFF
--- a/lib/services/groups/groupService.js
+++ b/lib/services/groups/groupService.js
@@ -27,6 +27,7 @@ var async = require('async'),
     iotManagerService = require('./../common/iotManagerService'),
     intoTrans = require('../common/domain').intoTrans,
     apply = async.apply,
+    deviceService = require('../devices/deviceService'),
     logger = require('logops'),
     config = require('../../commonConfig'),
     errors = require('../../errors'),
@@ -166,15 +167,38 @@ function handleWithIotaRegistration(callback) {
  * @param {String} resource             Resource where the devices will arrive.
  * @param {String} apikey               Api keys the devices will declare.
  */
-function remove(service, subservice, resource, apikey, callback) {
+function remove(service, subservice, resource, apikey, device, callback) {
     function extractId(deviceGroup, callback) {
         callback(null, deviceGroup._id);
+    }
+
+    function deleteDevices(device, service, subservice, id, callback) {
+        if(device) {
+            deviceService.listDevices(service, subservice, function (error, devices) {
+                if (error) {
+                   callback(error);
+                } else {
+                    if (devices && devices.count > 0){
+                      for(var device of devices.devices) {
+                        deviceService.unregister(device.id, service, subservice, function(error) {
+                            if (error) {
+                                callback(error);
+                            }
+                        });
+                      }
+                    }
+                }
+            });
+        }
+
+        callback(null, id);
     }
 
     async.waterfall([
         apply(config.getGroupRegistry().get, resource, apikey),
         apply(checkServiceIdentity, service, subservice),
         extractId,
+        apply(deleteDevices, device, service, subservice),
         config.getGroupRegistry().remove
     ], handleWithIotaRegistration(callback));
 }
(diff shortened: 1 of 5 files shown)


fgalan:
nock configuration could be tricky... if the request doesn't exactly match with what the mock expects, it fails without providing much more detail.

Anyway, in the case it helps, could you provide the exact ECONNECT error message you get?

chicco785:
I realised that it is a tricky tool, the weird thing in this case is that it looks like the agent cannot connect to the nock (I had before errors where the request was not matching and i understand the difference). Not sure there is a time out by which the nock is killed. I will re-run the test and pass the error.

fgalan:
> I will re-run the test and pass the error.

Not sure of understanding... you mean that the problem has been solved?

chicco785:
i just meant that i need to extract the full stack error and pass it here :) sorry for the confusion

chicco785:
88ce2be fixed the ECONNECT error, still to have the correct reply, i had to insert an empty call to the iot agent api, in the code, since (not sure why, maybe it was just a time issue) the registry delete operation was called after the call for checking the fact that the device where cancelled.

chicco785:
i believe this is ready for merge

lib/services/groups/groupService.js
```diff
@@ -27,6 +27,7 @@ var async = require('async'),
     iotManagerService = require('./../common/iotManagerService'),
```
fgalan: An entry in the CHANGES_NEXT_RELEASE regarding the changes introduced by this PR should be added (I guess that mentioning issue #596)

fgalan:
> i believe this is ready for merge

A comment regarding CHANGES_NEXT_RELEASE seems to be the only thing pending (although I'd have a new look at the end, given it has passed some time :) A sync updating recent master would be also a good idea.

chicco785:
Rebase for sure!

iFede - Fede's mobile edition

dcalvoalonso:
Just one comment, shall we include also some unit tests for NGSIv2?

chicco785:
currently device-api-group-test is not at all covered for NGSIv2 (i.e. also all the other tests have not been implemented for NGSIv2)

dcalvoalonso:
Maybe you have to do a pull of master branch, I see some NGSIv2 tests for provisioning: https://github.com/telefonicaid/iotagent-node-lib/tree/master/test/unit/ngsiv2/provisioning

chicco785:
hi @dcalvoalonso i had a quick look again, but it seems to me that none of the test covers device group provisioning.

dcalvoalonso:
Ups, you're right @chicco785 !! Sorry, too much GitHub issues for a single day!! :P

dcalvoalonso:
LGTM

fgalan:
@chicco785 sorry for the delay in this PR. Currently we are giving priority to a task related with upgrading node version in the library but once we finish with it, this PR will be progressed.

fgalan:
Merging. Thanks for the contribution!

Issue #577, opened 2018-02-08
Missing Releasing Alarm trace for MongoDB connection errors


fgalan:
It has been found (in PRO environment) that IOTAgents doesn't honour the Releasing alarm condition described here https://github.com/telefonicaid/iotagent-manager/blob/master/docs/operations.md#alarms in the case of MongoDB errors. Log trace:

```
time=2018-02-08T11:22:07.227Z | lvl=ERROR | corr=n/a | trans=n/a | op=IoTAgentNGSI.Alarms | msg=Raising [MONGO-ALARM]: undefined
{"time":"2018-02-08T11:33:11.452Z","lvl":"INFO","msg":"Setting IOTA_SERVER_PORT to environment value: 8076"}
{"time":"2018-02-08T11:33:11.455Z","lvl":"INFO","msg":"Setting IOTA_SERVER_HOST to environment value: 0.0.0.0"}
{"time":"2018-02-08T11:33:11.455Z","lvl":"INFO","msg":"Setting IOTA_LOG_LEVEL to environment value: INFO"}
time=2018-02-08T11:33:11.460Z | lvl=INFO | corr=n/a | trans=n/a | op=IoTAManager.NorthboundServer | msg=Starting IoT Agent Manager listening on port [8076]
time=2018-02-08T11:33:11.476Z | lvl=INFO | corr=n/a | trans=n/a | op=IoTAManager.DbConn | msg=First connection attempt
time=2018-02-08T11:33:11.477Z | lvl=INFO | corr=n/a | trans=n/a | op=IoTAManager.DbConn | msg=Attempting to connect to MongoDB instance. Attempt 1
time=2018-02-08T11:33:11.551Z | lvl=INFO | corr=n/a | trans=n/a | op=IoTAManager.DbConn | msg=Successfully connected to MongoDB.
time=2018-02-08T11:33:11.565Z | lvl=INFO | corr=n/a | trans=n/a | op=IoTAManager.Executable | msg=IoT Agent Manager started
```

Note the `[MONGO-ALARM]: undefined` at the begining but then MongoDB connection is restored (`Successfully connected to MongoDB`) without printing the corresponding Releasing message.

fgalan:
Maybe solved by PR https://github.com/telefonicaid/iotagent-node-lib/pull/605 ?

AlvaroVega:
Unfortunately no :(

fgalan:
[...]

PR #605, opened 2018-04-25
Update mongodb dependence in package.json

AlvaroVega:
Previous version (2.2.10) is already deprecated.
Maybe this update solves issue https://github.com/telefonicaid/iotagent-node-lib/issues/577


fgalan:
Check also packages.json in iotagent-ul and iotagent-json in order to remove mongodb packages if it is not actually needed (it should be so, as the agent should always access DB through the library and never directly)

PR #609, opened 2018-05-07
update to last stable mongodb and mongoose versions

AlvaroVega:


package.json
```diff
@@ -38,7 +38,7 @@
     "jison": "0.4.17",
```
fgalan: The following lines should be updated in CNR:

```
- Upgrade mongodb dependency version to 2.2.20
- Upgrade mongoose dependency version to 4.13.11
```

AlvaroVega: Fixed in https://github.com/telefonicaid/iotagent-node-lib/pull/609/commits/16a88a37a94ee6fbc8756bc1d3922185ec0363bc

PR #610, opened 2018-05-10
use retries and retry_time env vars

AlvaroVega:


lib/commonConfig.js
```diff
@@ -68,7 +68,9 @@ function processEnvironmentVariables() {
              'IOTA_MONGO_HOST',
```
fgalan: CNR about this change? Maybe something like this:

```
-Fix: IOTA_MONGO_RETRIES and IOTA_MONGO_RETRY_TIME env vars are now taken into account (previously there were ignored)
```

AlvaroVega: Fixed in https://github.com/telefonicaid/iotagent-node-lib/pull/610/commits/82abee64ad603f7874859360d316cd04c83a4313

Issue #604, opened 2018-04-13
[iot/services] Many groups, one servicePath 


gabrieledeluca:
**CONFIGURATION API** 

I have more of one group on the same servicePath. I can call POST _/iot/services_ on order to create an array of services on a _servicePath_ and, in fact, these are stored on mongodb (groups table) correctly. But when I call the GET service in order to fetch all services related to the _servicePath_, service return only one service. Why?

fgalan:
Maybe is a limitation of the current API

@dcalvoalonso what do you think?

gabrieledeluca:
Yes, the problem happens when I use mongodb (as deviceRegistry option). In order to clarify, below you can find a query result on my db and the service response associated.

*************************
**MONGODB** 
*************************
Query: **db.getCollection('groups').find({})**

/* 1 */
{
    "_id" : ObjectId("5afae83e73dcc47d6fc914ed"),
    "cbHost" : "192.168.56.101:1026",
    "subservice" : "/demo",
    "service" : "opcua_car",
    "type" : "Car",
    "apikey" : "801230BJKL23Y9090DSFL123HJK09H324HV8732",
    "resource" : "/Car",
    "attributes" : [ 
        {
            "type" : "Number",
            "name" : "Engine_Temperature"
        }, 
        {
            "type" : "Number",
            "name" : "Engine_Oxigen"
        }
    ],
    "lazy" : [ 
        {
            "type" : "Number",
            "name" : "Speed"
        }
    ],
    "staticAttributes" : [],
    "commands" : [ 
        {
            "type" : "command",
            "name" : "Stop"
        }, 
        {
            "type" : "command",
            "name" : "Accelerate"
        }
    ],
    "__v" : 0
}

/* 2 */
{
    "_id" : ObjectId("5afae8a24db94a7dab06648b"),
    "cbHost" : "192.168.56.101:1026",
    "subservice" : "/demo",
    "service" : "opcua_car",
    "type" : "Car2",
    "apikey" : "801230BJKL23Y9090DSFL123HJK09H324HV8732",
    "resource" : "/Car2",
    "attributes" : [ 
        {
            "type" : "Number",
            "name" : "Engine_Temperature"
        }, 
        {
            "type" : "Number",
            "name" : "Engine_Oxigen"
        }
    ],
    "lazy" : [ 
        {
            "type" : "Number",
            "name" : "Speed"
        }
    ],
    "staticAttributes" : [],
    "commands" : [ 
        {
[...]

PR #613, opened 2018-05-21
Repairs #604

dcalvoalonso:
This PR solves https://github.com/telefonicaid/iotagent-node-lib/issues/604.

It fixes some problems in groupService.js and groupRegistryMongoDB.js files.

The correctness of solution is demonstrated by adding a new test to mongodb-group-registry-test.js and modifying the expected value of another within this file.

Travis CI is passed and correct linting has been checked.

### DIFF
--- a/lib/services/groups/groupRegistryMongoDB.js
+++ b/lib/services/groups/groupRegistryMongoDB.js
@@ -160,16 +160,39 @@ function getById(id, callback) {
     });
 }
 
+function find(service, subservice, callback) {
+    var query;
+
+    function toObjectFn(obj) {
+        return obj.toObject();
+    }
+
+    logger.debug(context, 'Looking for device group with service [%s] and subservice [%s].', service, subservice);
+
+    query = Group.model.find({service: service, subservice: subservice});
+    query.select({__v: 0});
+
+    query.exec(function handleGet(error, data) {
+        if (error) {
+            logger.debug(context, 'Internal MongoDB Error getting device: %s', error);
+
+            callback(new errors.InternalDbError(error));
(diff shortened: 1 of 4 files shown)


lib/services/groups/groupRegistryMongoDB.js
```diff
@@ -160,6 +160,31 @@ function getById(id, callback) {
     });
 }
 
+function find(service, subservice, callback) {
```
fgalan: JavaDoc missing?

(The "?" is because I have doubts... on one hand no other function in this file includes JavaDoc, but maybe is a good idea to to at least for newly added functions)

lib/services/northBound/deviceGroupAdministrationServer.js
```diff
@@ -161,7 +161,7 @@ function handleListDeviceGroups(req, res, next) {
             req.query.offset,
             listHandler);
     } else {
-        groupService.find(req.headers['fiware-service'], req.headers['fiware-servicepath'], listHandler);
+        groupService.find(req.headers['fiware-service'], req.headers['fiware-servicepath'], null, listHandler);
```
AlvaroVega: What is supposed to be in ```null``` place?

dcalvoalonso: As you can see in https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/groups/groupService.js#L110, the third argument is the type, which is optional and not used by handleListDeviceGroups function.

You can see an example of using this function with type parameter in https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/groups/groupService.js#L257

PR #620, opened 2018-06-08
Improve readability of readthedocs documentation.

jason-fox:
* The link to readthedocs on the main readme has been corrected
* Readthedocs is unable to handle `<a href="xx">` anchor tags, this also messes up the sidebar in readthedocs. These have been removed.
* Table of Contents (ToC) has been regenerated for each documentation file.
* Readthedocs cannot handle fenced code blocks - an extra linefeed has been added where necessary.

To be honest the ToC is both confusing and unnecessary for the readthedocs documentation - it is generated on the sidebar automatically. Second level headings are also not indented by readthedocs

My preference would be to remove the extra table of contents on each readthedocs, but I have done a 1-to-1 replacement for now, therefore some pages have a ToC, others do not. Some have a one level ToC, others a 2 level ToC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
   + [Description](#description)
   + [Device to NGSI Mapping](#device-to-ngsi-mapping)
   + [Features](#features)
-  + [The ´TimeInstant´ element](#the-%C2%B4timeinstant%C2%B4-element)
+  + [The `TimeInstant` element](#the-timeinstant-element)
   + [Implementation decisions](#implementation-decisions)
 * [Build and Install](#build-and-install)
 * [API Overview](#api-overview)
@@ -235,7 +235,7 @@ to a command update to the device.
 
 ![General ](https://raw.githubusercontent.com/telefonicaid/iotagent-node-lib/master/img/iotAgentLib.png "Architecture Overview")
 
-### The ´TimeInstant´ element
+### The `TimeInstant` element
 
 As part of the device to entity mapping process the IoT Agent creates and updates automatically a special timestamp.
 This timestamp is represented as two different properties of the mapped entity::
@@ -1060,7 +1060,7 @@ The command line testing tools make use of the [command-node Node.js library](ht
 utils. In order to help creating testing tools for IoTAgents of specific protocols, all the commands of the library tester are offered as a array
 that can be directly imported into other Command Line tools, using the following steps:
 
-* Require the ´iotagent-node-lib´ command line module in your command line tool:
+* Require the `iotagent-node-lib` command line module in your command line tool:
 ```
   var iotaCommands = require('iotagent-node-lib').commandLine;
 ```
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 ## Index
 
 * [Overview](#overview)
-* [Build & Install](#build--install)
+* [Build and Install](#build-and-install)
 * [API Overview](#api-overview)
   + [About API](#about-api)
   + [Device Provisioning API](#device-provisioning-api)
--- a/README.md
+++ b/README.md
@@ -7,10 +7,16 @@
 ## Index
 
 * [Overview](#overview)
+  + [Description](#description)
+  + [Device to NGSI Mapping](#device-to-ngsi-mapping)
+  + [Features](#features)
+  + [The ´TimeInstant´ element](#the-%C2%B4timeinstant%C2%B4-element)
+  + [Implementation decisions](#implementation-decisions)
 * [Build and Install](#build-and-install)
 * [API Overview](#api-overview)
   + [About API](#about-api)
   + [Device Provisioning API](#device-provisioning-api)
+  + [Configuration API](#configuration-api)
 * [Advanced Topics](#advanced-topics)
   + [Secured access to the Context Broker](#secured-access-to-the-context-broker)
   + [Data mapping plugins](#data-mapping-plugins)
@@ -18,6 +24,7 @@
 * [Testing](#testing)
   + [Agent Console](#agent-console)
   + [Agent tester](#agent-tester)
+* [Development documentation](#-development-documentation)
 
 ## Overview
 ### Description
--- a/README.md
+++ b/README.md
@@ -1,25 +1,25 @@
 # FIWARE IoT Agent Framework
 
 [![License badge](https://img.shields.io/badge/license-AGPL-blue.svg)](https://opensource.org/licenses/AGPL-3.0)
-[![Documentation badge](https://readthedocs.org/projects/fiware-iotagent-node-lib/badge/?version=latest)](http://fiware-iotagent-ul.readthedocs.org/en/latest/?badge=latest)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -228,7 +228,7 @@ to a command update to the device.
 
 ![General ](https://raw.githubusercontent.com/telefonicaid/iotagent-node-lib/master/img/iotAgentLib.png "Architecture Overview")
 
-### <a name="TimeInstant"/>The ´TimeInstant´ element
+### The ´TimeInstant´ element
```
fgalan: Looking in the Orion documentation it seems that `<a name=...>` works ok. For intance, in https://raw.githubusercontent.com/telefonicaid/fiware-orion/master/doc/manuals/admin/metrics_api.md we have:

```
# <a name="top"></a>Metrics API
```

and in the RTD rendering of that document (at https://fiware-orion.readthedocs.io/en/master/admin/metrics_api/index.html) it looks fine:

imagen

and the links with "Top" also seem to work fine.

So, could you elaborate on the problem with anchors in this repo, please? Maybe is a matter of chaning `<a name=.../>` by the one seems to work for Orion repo, i.e. `<a name=...></a> (althouth from a HTML point of view is the same, maybe something in RTD causes problems with the former).

README.md
```diff
@@ -1,25 +1,25 @@
...
-  * [Configuration API](#configurationapi)
-* [Advanced Topics](#advancedtopics)
-  * [Secured access to the Context Broker](#securedaccess)
-  * [Data mapping plugins](#datamapping)
-  * [Old IoTAgent data migration](#datamigration)
-* [Testing](#librarytesting)
-* [Development Documentation](#development)
-
-## <a name="overview"/> Overview
+* [Build & Install](#build--install)
```
fgalan: If at the end

```
##  Build and Install
```

is used, then I undertand in the index we should have:

```
* [Build and Install](#build-and-install)
```

README.md
jason-fox: I agree the HTML rendering **should** be the same, except that it isn't. [sigh]

Go to https://iotagent-node-lib.readthedocs.io/en/latest/usermanual/index.html

Hover over the line:
>The library provides a mechanism for the periodic reporting of stats related to the library's work.
>In order to activate the use of the periodic stats, it must be configured in the config file, as 
>described in the Configuration section.
>

The **entire paragraph** appears to be a link. This is due to readthedocs mis-interpreting:

```html
### <a name="statsregistry"/> Stats Registry
```

in the line before as an open `<a href>` anchor tag

Of course the rendering of GFM on a Markdown file on **GitHub** itself is fine - it's readthedocs that can't handle it.

The rendering of the sidebar nav is also completely screwed up as well.

I would argue that in general  adding HTML `<a href>`  tags into Markdown files (although technically acceptable within the specifications most Markdown flavors) are:

* A Totally unnecessary hack (since Markdown has its own linking mechanism)
* More difficult to maintain than generating markdown style links
* Make the Raw Markdown file more difficult to read.

The main acceptable usage for  `<a href>` tags is for markdown files written in extended character sets (e.g. CJK) where the markdown ascii character linking just doesn't work. In general titles for files written in English won't have these issues.

README.md
jason-fox: Fixed in c80883

jason-fox: Actually  I  just regenerated the whole ToC using markdown-toc instead.  fixed in ad029d4
This will ensure all ToC links are valid without making manual edits.

aarranz:
Hi,

In WireCloud, we collected some useful plugins for Python Markdown (and therefore for MkDocs) that make it behave more like GFM. The package currently is called [wirecloud-markdown-github](https://github.com/Wirecloud/wirecloud_markdown_github). This package was created before the PyMdown extensions were published on pypi, so probably it is not needed anymore and we can use the PyMdown extensions directly. The only extension that I'm missing is the extension for making Python Markdown use the same section anchors ids that the ones created on GitHub.

To use markdowns plugins on Read The Docs, you have to create a [mkdocs_requirements.txt](https://github.com/Wirecloud/wirecloud/blob/develop/mkdocs_requirements.txt) file and enabling them on the [mkdocs.yml](https://github.com/Wirecloud/wirecloud/blob/develop/mkdocs.yml#L31) configuration file.

README.md
fgalan: I mostly agree, but not sure if somebody is using these anchors (e.g. in direct links to documentation in other places)... should that people get fails after the "cleanup"?

Maybe we can use this PR to change `<a name="..." />` to `<a name="..."></a> (as in Orion RTD rendering it seems to work) and see how it renders in RTD. If after doing that it is still having problem, we can remove `<a>` usages completely.

What do you think?

jason-fox: My opinion, as stated above - `<a name="."/>` should be **deleted** - easier to read, easier to maintain.

Since this document wasn't even on readthedocs until two weeks ago, I doubt there are any **external** deep links. Even then, usually it would make more sense to deep link to the head of each file if a deep-link is necessary e.g. https://iotagent-node-lib.readthedocs.io/en/latest/usermanual/ 

So the only likely broken links will be found within the document itself. The ToC has been regenerated so those links should work, so it can be reduced to internal links within the body of the document. Personally if I come across a dead link in the body of the document I'd probably click on the sidebar navigation (which removing `<a name="."/>`  tag fixes in readthedocs anyway)

Besides which, a failed deep link will also display the head of the document e.g.: https://iotagent-node-lib.readthedocs.io/en/latest/usermanual/#x - at worst this will be an `<h1>` rather than an `<h3>` 

Finally,  if we leave `<a name="."/>`  tags in there you are only encouraging someone else to start using them.

I say **delete** :smiley:

fgalan: The document wasn't even on readthedocs until two weeks ago but it has been time along in markdown. Thus, people using links like

https://github.com/jason-fox/iotagent-node-lib/blob/master/README.md#TimeInstant

will have them broken after merging PR.

However, I think is a good "hygienic" change the two ways of linking. So people that get the above link broken (not so broken, as at the end, they will land at the top of README.md) could change to the right one according to markdown internal way of linking, i.e:

https://github.com/jason-fox/iotagent-node-lib/blob/master/README.md#the-timeinstant-element

Thus, NTC :)

README.md
```diff
@@ -1,25 +1,32 @@
...
-  * [Data mapping plugins](#datamapping)
-  * [Old IoTAgent data migration](#datamigration)
-* [Testing](#librarytesting)
-* [Development Documentation](#development)
-
-## <a name="overview"/> Overview
+  + [Description](#description)
+  + [Device to NGSI Mapping](#device-to-ngsi-mapping)
+  + [Features](#features)
+  + [The ´TimeInstant´ element](#the-%C2%B4timeinstant%C2%B4-element)
```
fgalan: Broken link. It seems that

```
#the-timeinstant-element
```

solves it.

In addition, maybe `´TimeInstant´` should be changed to 

```
`TimeInstant`
```

I guess that the original idea was to use "code style" for that, but probably the author use the wrong symbol :)

jason-fox: Fixed in 5e4e34e

Issue #621, opened 2018-06-12
timestampProcessPlugin not allow multiEntity plugin results in NGSIv2


AlvaroVega:
Related with PR about multientity and NGSIv2: https://github.com/dcalvoalonso/iotagent-node-lib/pull/5

If config.iota.timestamp flag is active then timestampProcessPlugin is executing before multiEntity plugin. Multientity plugin expands payload json in an array of entities. But TimeStampProcess plugin only expects a entity, and transformation performed by timestampProcess generates a wrong payload.

Here https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/ngsi/ngsiService.js#L435 result is not an array if timeStampPlugin is active.

Here https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/plugins/timestampProcessPlugin.js#L59 entity is an array of entites when multientity is preset

To reproduce it, add config.iota.timestamp flag to 
https://github.com/dcalvoalonso/iotagent-node-lib/pull/5/files#diff-8f552cb201d86830244748779fd2ee9bR35

dcalvoalonso:
Sure @AlvaroVega! I will check it right now and provide a solution ASAP

AlvaroVega:
Example:

$ curl -i -X GET 'http://localhost:7896/iot/d?i=disp1&d=t|18&k=vfd5f8n3ry0mk0vwnlr0niopw'

Produces a logs like:
```
iot-iota-ul                 | time=2018-06-12T10:33:40.787Z | lvl=DEBUG | corr=0eaeb863-2453-4913-b8d1-dd7c716d04aa | trans=0eaeb863-2453-4913-b8d1-dd7c716d04aa | op=IOTAUL.HTTP.Binding | srv=smartcity | subsrv=/ | msg=Parsing payload [t|18] | comp=IoTAgent
iot-iota-ul                 | time=2018-06-12T10:33:40.787Z | lvl=DEBUG | corr=0eaeb863-2453-4913-b8d1-dd7c716d04aa | trans=0eaeb863-2453-4913-b8d1-dd7c716d04aa | op=IOTAUL.HTTP.Binding | srv=smartcity | subsrv=/ | msg=Parsed data: [[{"t":"18"}]] | comp=IoTAgent
[...]

PR #623, opened 2018-06-12
Fixes bug #621 to support timestampProcessPlugin to be executed after multientity plugin

dcalvoalonso:
This PR solves issue #621. The following changes have been done:
- New test added to detect the bug.
- Modification of timestampProcessPlugin to check if the input is an array and apply the functionality using map.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -17,3 +17,4 @@
 - Add: configuration service is exposed as part of the module
 - Add: casting attribute's values to JSON native types (activable by config.autocast)
 - Fix: GET call to northbound group endpoint returns multiple services (#604)
+- Fix: timestampProcessPlugin supports processing an array of NGSI entities (#621)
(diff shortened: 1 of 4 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -17,3 +17,4 @@
 - Add: configuration service is exposed as part of the module
 - Add: casting attribute's values to JSON native types (activable by config.autocast)
 - Fix: GET call to northbound group endpoint returns multiple services (#604)
+- Fix: timestampProcessPlugin supports processing an array of NGSI entities (#621)
```
fgalan: Maybe it would be more precise?

> - Fix: timestampProcessPlugin supports processing an array of NGSI entities in NGSIv2 (#621)

(Of course, if I'm right -I'm not sure- and the problem only happens with NGSIv2. It could be considered part of the already existing "Add: allow NGSIv2 for updating active attributes..." but better to be explicit, so good idea to include a new line for this PR)

dcalvoalonso: I agree with you, it's better to have this entry in the CR as a separate line to improve traceability.

dcalvoalonso: Fixed in d0fafa981fb7f1363e8110b6483c0c0b8f7a7f33

fgalan: Typo: NGSIc2

dcalvoalonso: Fixed in e9b0fdd8a82a353db7d0e6c4d52617d20fad1c96

PR #626, opened 2018-06-14
fix multientity attributes duplicated in entity

AlvaroVega:
Fix: https://github.com/telefonicaid/iotagent-node-lib/issues/627


manucarrace:
LGTM

PR #642, opened 2018-06-29
(reopen) replace undefined by current alarm description: print alarm description as object

AlvaroVega:
Reopen previous PR https://github.com/telefonicaid/iotagent-node-lib/pull/640/

### DIFF
--- a/lib/services/common/alarmManagement.js
+++ b/lib/services/common/alarmManagement.js
@@ -41,7 +41,7 @@ function raise(alarmName, description) {
             description: description
         };
 
-        logger.error(context, 'Raising [%s]: %s', alarmName, description);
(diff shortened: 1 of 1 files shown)


lib/services/common/alarmManagement.js
```diff
@@ -41,7 +41,7 @@ function raise(alarmName, description) {
             description: description
         };
 
-        logger.error(context, 'Raising [%s]: %s', alarmName, description);
+        logger.error(context, 'Raising [%s]: %j', alarmName, description);
```
fgalan: What's the difference between `%s` and `%j`?

(Just curiosity... if some day I have known it, I have forgot :)

fgalan: (The doubt is in `%j`... `%s` I guess stand for string)

PR #645, opened 2018-07-17
Clarified North/South, removed anchors, added code highlights

jason-fox:
After discussion with @dcalvoalonso 

With standard English usage, the “bound”  suffix in implies a direction e.g “Homeward bound" means "going **towards** my home”, therefore “Southbound” means "going **towards** the South”
In contrast, the phrase "to the North of” does not imply direction, it implies relative location — you can be to the North of a town, but travelling Southbound - i.e. towards the city.
The main common usage of Northbound/Southbound is found with motorway traffic reports e.g. “the Northbound carriageway is blocked"

Therefore the following usage is correct common English, and the existing documents should be corrected to use it: 

* Traffic **to the South of** the IoT Agent refers to the exchange of information and commands between a device and an IoT Agent (Native protocols)
Traffic **to the North of** the IoT Agent traffic considers the link between the IoT Agent  and the CB.  (NGSI)
* **Southbound** traffic refers to the commands sent from the CB down to the device
* **Northbound** traffic refers to the measurements sent up from the IoT device to the CB
* **Northbound** traffic also refers to the results of commands  sent up from the IoT device to the CB

I have updated the documents  and removed `<a>`  tags and added code highlights for ReadTheDocs whilst updating. (RTD doesn't support `console`)

### DIFF
--- a/doc/howto.md
+++ b/doc/howto.md
@@ -282,7 +282,7 @@ In order to do so, add the following require statement to the initialiation code
 
 and add the `request` dependency to the `package.json` file:
 
-```json
+```
   "dependencies": [
   [...]
 
--- a/doc/howto.md
+++ b/doc/howto.md
@@ -14,7 +14,7 @@
   + [Provisioning handlers](#provisioning-handlers)
 
 ## Overview
-This document's goal is to show how to develop a new IOT Agent step by step. To do so, a simple invented HTTP protocol
+This document's goal is to show how to develop a new IoT Agent step by step. To do so, a simple invented HTTP protocol
 will be used, so it can be tested with simple command line instructions as `curl` and `nc`.
 
 ### Protocol
@@ -33,10 +33,13 @@ Where:
 This tutorial expects a Node.js v0.10 (at least) installed and working on your machine. It also expects you to have
 access to a Context Broker (without any security proxies).
 
-##  Basic IOTA
-In this first chapter, we will develop an IOT Agent with a fully working NGSI traffic on the North port and no native
-traffic south of the IoT Agent. This may seem useless (and indeed it is) but will serve us well on showing the basic
-steps in the creation of an IoT Agent.
+##  Basic IoT Agent
+
+In this first chapter, we will just develop an IoT Agent with a fully connected North Port. This will
+send and receive NSGI traffic and can be administered using the IoT Agent's Device Provisioning
+API. The South Port will remain unconnected and no native protocol traffic will be sent to the devices.
+This may seem useless (and indeed it is) but it will serve us well on showing the basic steps in the creation
+of an IoT Agent.
 
 First of all, we have to create the Node project. Create a folder to hold your project and type the following
 instruction:
@@ -64,7 +67,7 @@ The first step is to write a configuration file, that will be used to tune the b
 can be copied from the `config-basic-example.js` file, in this same folder. Create a `config.js` file with it
 in the root folder of your project. Remember to change the Context Broker IP to your local Context Broker.
 
-Now we can begin with the code of our IOTA. The very minimum code we need to start an IOTA is the following:
+Now we can begin with the code of our IoT Agent. The very minimum code we need to start an IoT Agent is the following:
 
 ```javascript
 var iotAgentLib = require('iotagent-node-lib'),
@@ -78,7 +81,7 @@ iotAgentLib.activate(config, function(error) {
 });
 ```
 
-The IOTA is now ready to be used. Execute it with the following command:
+The IoT Agent is now ready to be used. Execute it with the following command:
 
 ```bash
 node index.js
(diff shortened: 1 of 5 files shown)


doc/howto.md
```diff
@@ -280,15 +280,15 @@ In order to do so, add the following require statement to the initialiation code
     request = require('request');
 ```
 
-and add the `request` dependency to the `package.json` file: 
+and add the `request` dependency to the `package.json` file:
 
 ```json
```
fgalan: This is similar to https://github.com/telefonicaid/lightweightm2m-iotagent/pull/130#discussion_r203048085

In this case, given this is a fragmetn of a JSON which doesn't striclty follows the JSON syntax I'd suggest don't add the highligth keyword.

jason-fox: Fixed 3c19e67 - Removed Highlight

jason-fox:
We probably should wait for ReadtheDocs to sort out their build issue before merging
see: https://github.com/rtfd/readthedocs.org/issues/4383

fgalan:
mmm... I have also recieved a couple of "Build Failed" emails from RTD in another repo (fiware-orion). Maybe they are having problems today. Yes, let's wait gets fixed before merging.

jason-fox:
It looks like RTD have fixed the build issue now.

doc/howto.md
```diff
@@ -34,14 +34,14 @@ This tutorial expects a Node.js v0.10 (at least) installed and working on your m
 access to a Context Broker (without any security proxies).
 
 ##  Basic IOTA
-In this first chapter, we will develop an IOT Agent with a fully working northbound API and no southbound
-API. This may seem useless (and indeed it is) but will serve us well on showing the basic steps in the creation
-of an IOTA.
- 
-First of all, we have to create the Node project. Create a folder to hold your project and type the following 
+In this first chapter, we will develop an IOT Agent with a fully working NGSI traffic on the North port and no native
```
dcalvoalonso: Of course, I don't want to make English suggestions to you @jason-fox but this sentence is a bit strange... I mean, maybe it would be better something like _the IoT will have a fully working NGSI north API and no IoT native south API_. I think that we must highlight the term **API**.

dcalvoalonso: Looking at other changes you have introduced, maybe we could use The **North interface** or something similar.

doc/howto.md
```diff
@@ -134,8 +134,8 @@ function initSouthbound(callback) {
 }
 ```
 
-This Express code sets up a HTTP server, listening in the 8080 port, that will handle incoming requests targeting 
-path `/iot/d` using the middleware `manageULRequest()`. This middleware will contain all the Southbound logic, and 
+This Express code sets up a HTTP server, listening in the 8080 port, that will handle incoming requests targeting
+path `/iot/d` using the middleware `manageULRequest()`. This middleware will contain all the Southbound logic, and
```
dcalvoalonso: This middleware will contain all the Southbound logic?

doc/usermanual.md
```diff
@@ -83,7 +83,7 @@ Returns the current log level, in a json payload with a single attribute `level`
 
 ### Transactions
 The library implements a concept of transactions, in order to follow the execution flow the library follows when treating
-requests entering both from the Northbound and the Southbound.
+requests entering both from the North and the South ports of the IoT Agent.
```
dcalvoalonso: In line 90:

> Southbound requests, the user is responsible of

doc/northboundinteractions.md
```diff
@@ -1,4 +1,4 @@
-# Northbound NGSI Interactions
+# Traffic North of the IoT Agent -  NGSI Interactions
```
dcalvoalonso: In line 60:

> South Bound protocols (typically lightweight protocols aimed to constrained devices)

dcalvoalonso: In line 65:

> Devices, in the South Bound

doc/usermanual.md
```diff
@@ -83,7 +83,7 @@ Returns the current log level, in a json payload with a single attribute `level`
...
 
 To follow the transactions, a new Domain is created for each incoming request; in the case of Northbound requests, this
```
dcalvoalonso: in the case of Northbound requests

doc/usermanual.md
dcalvoalonso: In lines 521:

> This collection of utility middlewares is aimed to be used in the northbound of the IoTAgent Library

fgalan:
> Just a final thought, in the case of this project, specially northbound term is widely used in the code. Shall we open a new issue and PR to refactor this? @fgalan @jason-fox

In my opinion, code comments are important but less important in terms of language polishing than documentation... For me not very important, but I wouldn't oppose if somebody want to create the issue and fix it :)

fgalan:
... but I realize that maybe you refer not only to code comments, but also literals (variables, etc.) and log traces. With regards to literals, it is also fine (at the end, they are "internal" to the code and shouldn't have effect to usrs). With regards to traces, debug level traces shouldn't be a problem (would the change affect to info, warn and error traces?)

doc/howto.md
jason-fox: Fixed - 376c3e2 - Replaced with *logic south of ...*

doc/usermanual.md
jason-fox: Fixed  376c3e2 - replaced with *requests received on the South Port*

doc/northboundinteractions.md
jason-fox: fixed  376c3e2 -  Replaced with *traffic sent and received on the South Port*

doc/usermanual.md
jason-fox: Fixed 376c3e2  Replaced with *requests received on the North Port of the IoT Agent*

doc/usermanual.md
jason-fox: Fixed a7e6db8 - Replace with *to the north of*

doc/howto.md
jason-fox: Its good to get a review - that sentence **was** incorrect - I introduced a typo.
Fixed - 376c3e2 - Replaced with *... an IoT Agent with a fully connected North Port. This will
send and receive NSGI traffic and can be administered using the IoT Agent's Device Provisioning API. The South Port will remain unconnected and no native protocol traffic will be sent to the devices*

dcalvoalonso:
@fgalan , I have checked traces and they don't include `northbound` or `southbound` terms. I was thinking more about variables but if you think that it is OK, I don't see either more problems.

dcalvoalonso:
LGTM!

jason-fox:
@fgalan - could you give the [readthedocs](https://readthedocs.org/projects/iotagent-node-lib/) build another kick? I looks like the githook is missing or hasn't triggered properly.

It might make sense to broaden the list of docs maintainers as well - at the moment you are the only one who can force a new build.

fgalan:
Re-build launched!

Issue #652, opened 2018-08-20
Missed protocol in device if was not provided in body at provision (but as url param)


AlvaroVega:
Trying to provision a device without protocol in body (just in url as a param):
```
curl -X POST \
  'https://195.235.93.235:8088/iot/devices?protocol=IoTA-JSON' \
  -H 'Cache-Control: no-cache' \
  -H 'Content-Type: application/json' \
  -H 'Postman-Token: 345f7fe5-8849-4615-bb59-f07ff84caa1c' \
  -H 'fiware-service: demo' \
  -H 'fiware-servicePath: /' \
  -H 'x-auth-token: 443bb86af4b94612b6657974b38aaf06' \
  -d '{
    "devices": [{
        "device_id": "electrovalve205",
        "entity_name": "electrovalve205",
        "entity_type": "electrovalve",
        "transport": "MQTT",
        "timezone": "Europe/Madrid",
        "attributes": [{
            "object_id": "t",
            "name": "temperature",
            "type": "int"
        }],
        "commands": [{
            "name": "activate",
            "type": "command"
        }],
        "static_attributes": [{
            "name": "att_name",
            "type": "string",
            "value": "value"
        }]
    }]
}'
```
Device is provisioned without protocol in body: 
```
{"devices":[{"device_id":"electrovalve216","entity_name":"electrovalve216","entity_type":"electrovalve","transport":"MQTT","timezone":"Europe/Madrid","attributes":[{"object_id":"t","name":"temperature","type":"int"}],"commands":[{"name":"activate","type":"command"}],"static_attributes":[{"name":"att_name","type":"string","value":"value"}]}]}
```
POSSIBLE FIX: Complete protocol in device provision if no provided since each agent know its protocol

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/653

PR #653, opened 2018-08-20
Ensure protocol in device provision

AlvaroVega:
Fix issue https://github.com/telefonicaid/iotagent-node-lib/issues/652

### DIFF
--- a/lib/services/devices/deviceRegistryMongoDB.js
+++ b/lib/services/devices/deviceRegistryMongoDB.js
@@ -23,6 +23,7 @@
 
 var logger = require('logops'),
     dbService = require('../../model/dbConn'),
+    config = require('../../commonConfig'),
     fillService = require('./../common/domain').fillService,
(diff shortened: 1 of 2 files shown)


lib/services/devices/deviceRegistryMongoDB.js
```diff
@@ -66,6 +67,11 @@ function storeDevice(newDevice, callback) {
         deviceObj[attributeList[i]] = newDevice[attributeList[i]];
     }
 
+    // Ensure protocol is in json
```
fgalan: This is generic code (in the library) independent from transport (JSON, UL or whatever... so this comment text is a bit surprising. Maybe the following is more accurate?

```
// If protocol is missing, then use default one from configuration
```

althougth I could be interpreting incorrectly the fix in this PR

AlvaroVega: protocol, is expected to be in newDevice object

Issue #663, opened 2018-09-18
Support authentication with Oauth2 using FIWARE Keyrock IDM


dcalvoalonso:
As it is explained in https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/installationguide.md, for the current implementation of OAuth2 based authentication:

> The code was tested using Keycloak and Auth0 (it may require customisation for other providers - while OAuth2 is a standard, not all implementations behave in the same way, especially as regards status codes and error messages.

Therefore, in order to support [FIWARE IDM Keyrock](https://github.com/ging/fiware-idm), we need to take into account the responses when an access token is asked.

In addition, the current implementation only covers the case of an IoT Agent sending updates to the CB, but the operations derived of a device provisioning request are not included.

fgalan:
Implemented by PR #664

fgalan:
PR #664 doesn't cover NGSIv2 support. It will be done in a separate PR. See https://github.com/telefonicaid/iotagent-node-lib/pull/664#discussion_r218820849

fgalan:
After PR #664, PR https://github.com/telefonicaid/iotagent-node-lib/pull/726 lands the contribution into master.

fgalan:
Now that PR #726 has been merged, should this issue be closed or is there something pending?

PR #664, opened 2018-09-18
Support authentication with Oauth2 using FIWARE Keyrock IDM

dcalvoalonso:
This PR aims to solve https://github.com/telefonicaid/iotagent-node-lib/issues/663.

Unit tests are added to validate the implementation.

It should be extended to supports NGSIv2 use.


lib/services/common/securityServiceOAuth2.js
```diff
@@ -21,6 +21,8 @@
  * For those usages not covered by the GNU Affero General Public License
```
fgalan: I undesrtand one entry (or more if needed) should be added to CHANGE_NEXT_RELEASE file, corrresponding to the changes included in this PR.

lib/services/common/securityServiceOAuth2.js
```diff
@@ -21,6 +21,8 @@
...
  * please contact with::daniel.moranjimenez@telefonica.com
  * This code was contributed by Federico M. Facca on behalf of Martel Innovate
```
fgalan: Should this PR involves changes in configuration or documentation?

test/unit/general/contextBrokerOAuthSecurityAccess-test.js
```diff
@@ -19,6 +19,8 @@
  *
```
fgalan: Are NGSIv2 unit test planed for this PR? Or they will be implemetented in a next one (also part of #663)?

fgalan:
We have some pending PRs related with auth. In particular: https://github.com/telefonicaid/iotagent-node-lib/pull/600 and https://github.com/telefonicaid/iotagent-node-lib/pull/603.

Is the work done in this PR and independent piece of work? Or it is related in some way?

CC: @chicco785

chicco785:
they should be related i suppose :) but i have to look in the code

lib/services/common/securityServiceOAuth2.js
```diff
@@ -64,19 +66,29 @@ function auth(trust, callback) {
...
             logger.debug(context, 'Authentication with trust [%s] was succesful', trust);
-
+            logger.debug(context,
+                'Received the following response from the OAuth2 provider:\n\n%s\n\n', JSON.stringify(body, null, 4));
             callback(null, response);
         } else if (response.statusCode === 401 || (response.statusCode === 400 &&
                 JSON.parse(response.body).error_description === 'Invalid refresh token')) {
             logger.error(context, 'Authentication rejected: %s', trust);
             callback(new errors.AuthenticationError(trust));
+        } else if (response.statusCode === 500 &&
```
chicco785: eh... this is not what the standard expect... i wonder if we should (for matter of improving the implementation of the standard) inform keyrock team about that.

chicco785: https://tools.ietf.org/html/rfc6749#section-5.2

lib/services/common/securityServiceOAuth2.js
```diff
@@ -64,19 +66,29 @@ function auth(trust, callback) {
...
 
     logger.debug(context, 'Authentication on the OAuth2 provider [%s] using trust [%s]', url, trust);
+    logger.debug(context, 'Using the following request:\n\n%s\n\n', JSON.stringify(options, null, 4));
 
     request(options, function handleResponse(error, response, body) {
         if (error) {
             logger.error(context, 'OAUTH2-001: Error retrieving token from OAuth2 provider: %s', error);
             callback(new errors.TokenRetrievalError(trust, error));
-        } else if (response.statusCode === 201) {
+        } else if (response.statusCode === 201 || response.statusCode === 200 /* Keyrock response */) {
```
chicco785: in this case it looks that the keycloak and auth are using wrong code:
https://tools.ietf.org/html/rfc6749#section-5.1

lib/services/common/securityServiceOAuth2.js
```diff
@@ -64,19 +66,29 @@ function auth(trust, callback) {
...
         } else if (response.statusCode === 401 || (response.statusCode === 400 &&
                 JSON.parse(response.body).error_description === 'Invalid refresh token')) {
             logger.error(context, 'Authentication rejected: %s', trust);
             callback(new errors.AuthenticationError(trust));
+        } else if (response.statusCode === 500 &&
+                JSON.parse(response.body).name === 'server_error') { //Keyrock response
+            logger.error(context, 'Authentication rejected: %s', trust);
+            callback(new errors.AuthenticationError(trust));
+        } else if (response.statusCode === 500 &&
+                JSON.parse(response.body).name === 'invalid_grant') { //Keyrock response
```
chicco785: https://tools.ietf.org/html/rfc6749#section-5.2

chicco785:
also on a side note, some oauth 2 servers are able to provide never expiring tokens (unless you revoke them), this simplify more the interaction (because you don't need to invoke at all then the oauth endpoint) and so the communication overhead, not sure this is the case for keyrock. still it could be good to introduce an option to suppor that.

lib/services/ngsi/ngsiService.js
```diff
@@ -690,7 +690,19 @@ function sendQueryValue(entityName, attributes, typeInformation, token, callback
  */
 
 function updateTrust(deviceGroup, deviceInformation, trust, response, callback) {
-    callback(null, response);
+    if (deviceGroup && response && response.body) {
```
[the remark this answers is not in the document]
chicco785: it could be at the level of the configuration of the agent e.g. `authenticationo.oauth_offline_token   = 1`

lib/services/common/securityServiceOAuth2.js
dcalvoalonso: Fixed with 7a19d2aaa796262b7eaefd113344d885838727f5

lib/services/common/securityServiceOAuth2.js
dcalvoalonso: Fixed with f55e7fb337fb4f14eb26f8c021aca3fe503cd44d

lib/services/common/securityServiceOAuth2.js
dcalvoalonso: OK. NTC

lib/services/common/securityServiceOAuth2.js
dcalvoalonso: I will send an email to Keyrock team (and open an issue in Keyrock GE). While they fix this issue, what do you think that we must do? Wait to the fix or go on with the PR?

lib/services/ngsi/ngsiService.js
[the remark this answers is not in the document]
dcalvoalonso: I implemented this because in my tests, each time I requested a new `access_token` to Keyrock using the `refresh_token`, the latter one became invalid. I will also ask this to Keyrock team.

lib/services/devices/deviceService.js
```diff
@@ -260,6 +261,51 @@ function formatCommandsNgsi2(originalVector) {
     return attributeList;
 }
 
+/**
+ * Executes a request operation using security information if available
```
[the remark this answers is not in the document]
dcalvoalonso: You're right, this PR covers also PR #603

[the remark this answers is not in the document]
dcalvoalonso: NTC

dcalvoalonso:
@chicco785 comment:
> so probably #603 is now replaced by this PR (if i got right, @dcalvoalonso ?) while subscriptions are not yet supported, which means #600 is still valid

Yes, I agree.

@chicco785  comment:
> also on a side note, some oauth 2 servers are able to provide never expiring tokens (unless you revoke them), this simplify more the interaction (because you don't need to invoke at all then the oauth endpoint) and so the communication overhead, not sure this is the case for keyrock. still it could be good to introduce an option to suppor that.

I will contact Keyrock team with these doubts and let's see what they think.

test/unit/general/contextBrokerOAuthSecurityAccess-test.js
dcalvoalonso: Yes, I plan to do a new PR as part of #663 but I preferred not to do this in a single request just for simplicity. If you agree, when we have the LGTM to this PR, I will do the new one with the NGSIv2 support.

lib/services/ngsi/ngsiService.js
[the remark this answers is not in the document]
dcalvoalonso: After checking the standard it says:

>    The authorization server MAY issue a new refresh token, in which case
   the client MUST discard the old refresh token and replace it with the
   new refresh token.  The authorization server MAY revoke the old
   refresh token after issuing a new refresh token to the client.  If a
   new refresh token is issued, the refresh token scope MUST be
   identical to that of the refresh token included by the client in the
   request.

So I understand that in this case, Keyrock implementation is right.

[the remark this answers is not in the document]
chicco785: hi @dcalvoalonso as mentioned in the email with upm team, this is the case if you use a "normal" refresh token and not the one obtained with 'offline_access' [scope] [1]. why it is better to use the 'offline_access' modality (i.e. a never expiring refresh token)?
in case, for whatever reason (no update for the sensor over a long time, broken connectivity, iot agent was down for some time), you don't refresh the refresh_token before its expiration, you would need to do again a manual authentication and update the refresh token in the configuration. 

[1]: https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess.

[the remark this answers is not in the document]
chicco785: see here some discussion on refresh token expiration: https://medium.com/@bantic/more-oauth-2-0-surprises-the-refresh-token-1831d71f4af6

lib/services/common/securityServiceOAuth2.js
chicco785: I think we don't need to wait their fix, we will eventually retro fit. so this is should not block the submission, while i think the support for the 'offline_token' (i.e. never expiring refresh token) should be added, it's not much work, 1st week of october i can fix that (if you don't want to integrated it here).

chicco785:
so the email, let's see replies, in general to reduce communication overhead, unless they have better (standard compliant...) ideas, i think the options we have are:

1. use directly a never expiring access token, that would be the simplest way. we could support this modality as an alternative to the refresh_token during device registration.

1. using the refresh token with offline_access (which is stored during the device configuration) and adding on top of that a caching mechanism to store the latest access token including its expiration, so that a new access token is requested only if the token is expired.

lib/services/common/securityServiceOAuth2.js
dcalvoalonso: Fixed with c7d9daed5a1ebcef4a33a6c3b07809a2b8cb9588

lib/services/common/securityServiceOAuth2.js
dcalvoalonso: Fixed with c7d9daed5a1ebcef4a33a6c3b07809a2b8cb9588

lib/services/ngsi/ngsiService.js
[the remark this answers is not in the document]
dcalvoalonso: Fixed with 2aab653c5bdcf5703ae470cd81925f7289ae4e45

test/unit/general/contextBrokerOAuthSecurityAccess-test.js
fgalan: Ok. Noted in #663.

NTC

dcalvoalonso: Finally, I decided to extend the tests for NGSIv2 in this PR.

dcalvoalonso: Fixed with a71f9fd745198cc4050d16984497db3fc6dc80f7

dcalvoalonso:
@fgalan @chicco785 @jason-fox I think that the PR is ready to get your final comments and review. Thanks in advance!

lib/commonConfig.js
```diff
@@ -65,6 +65,7 @@ function processEnvironmentVariables() {
              'IOTA_AUTH_PORT',
              'IOTA_AUTH_USER',
              'IOTA_AUTH_PASSWORD',
+             'IOTA_AUTH_PERMANENT_TOKEN',
```
fgalan: I understand the new configuration should be described at:

* https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/installationguide.md#global-configuration
* https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/installationguide.md#configuration-using-environment-variables

test/unit/examples/oauthRequests/getTokenFromTrustKeyrockGroup.json
```diff
@@ -0,0 +1 @@
+client_id=context-broker&client_secret=c8d58d16-0a42-400e-9765-f32e154a5a9e&grant_type=refresh_token&refresh_token=8970A9078A803H3BL98PINEQRW8342HBAMS
```
fgalan: I understand the tokens used in this an other examples in this PR ar fake, isn't it?

lib/commonConfig.js
dcalvoalonso: This was already done for https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/installationguide.md#global-configuration as you can see in https://github.com/telefonicaid/iotagent-node-lib/pull/664/files#diff-9b0a554150d1aeae259b560098da86f1

dcalvoalonso: Fixed with a50cc84a4aaa271e8dc09919c2c60d8b13be1e97

dcalvoalonso:
> @dcalvoalonso I have completed my review. Only some minor comments.
> 
> With regards to
> 
> > @dcalvoalonso  comment:
> > > @chicco785 comment:
> > > so probably #603 is now replaced by this PR (if i got right, @dcalvoalonso ?) while subscriptions are not yet supported, which means #600 is still valid
> > 
> > 
> > Yes, I agree.
> 
> If at the end this PR overspasses PR #603 I understand it should be closed. On the other and, I undesrtand that PR #600 will be rebased after this one and continue its way. Is that correct @chicco785 ?

I agree with your proposal @fgalan.

test/unit/examples/oauthRequests/getTokenFromTrustKeyrockGroup.json
fgalan: NTC

lib/commonConfig.js
fgalan: Commit a50cc84 includes the new env var in the env vars table but not the explanation of the new parameter (authentication.permanentToken), which is typically done in the just above section "Global Configuration".

Or maybe I'm missing something?

dcalvoalonso: This was done in https://github.com/telefonicaid/iotagent-node-lib/pull/664/commits/6efbf2f5b4df016d47f1ef6088765f4e07f20176

fgalan: Ok, here is the text:

> In order to use this type of token, an additional parameter `permanentToken` must be set to `true` in the `authentication` configuration. An environment variable `IOTA_AUTH_PERMANENT_TOKEN` can be also used for the same purpose

I missed it in the dense block of text in which it appears :)

NTC

fgalan:
@dcalvoalonso please solve the conflict with CHANGES_NEXT_RELEASE so this PR can be merged.

dcalvoalonso:
Conflicts solved with d90531d393db37f8c490e59ca78e40757a3a7e36

fgalan:
Merging to enable e2e integration pass with task/keyrockOauth2Integration-prelanding branch. @chicco785 and @jason-fox still pending, but they can be done in the PR that will follow this to merge task/keyrockOauth2Integration-prelanding into master

PR #670, opened 2018-09-26
Updates unit tests to work with newer version of nock

dcalvoalonso:


lib/plugins/attributeAlias.js
```diff
@@ -38,7 +38,10 @@ function extractSingleMapping(previous, current) {
 
     previous.direct[current.object_id] = current.name;
```
fgalan: An entry in the CNR file about this fix should be added

dcalvoalonso: Fixed in 06f744cfe57ae38e40bd57a84028f20a40bf0e53

CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
 Add protocol in device provision if not provided (#652)
-Fix: timestamp is now included just in non empty (more than id and type) multienty entitites
\ No newline at end of file
+Fix: timestamp is now included just in non empty (more than id and type) multienty entitites
+Fix: object_id fields are not introduced in CB requests by alias plugon (#660)
```
fgalan: plugon? :)

dcalvoalonso: Fixed with 2c67b6d7908ceae3e793ce993f99b73b678f3e14

Issue #672, opened 2018-09-27
TimeInstant must consider timezone attribute


dcalvoalonso:
With current implementation, TimeInstant values are always expressed using UTC timezone. The timezone attribute provided when a device or group is provisioned shall be used.

This has been requested by users:
https://stackoverflow.com/questions/35992228/fiware-idas-timeinstant-attribute
https://stackoverflow.com/questions/51428637/fiware-idas-ul-timeinstant-attribute

fgalan:
I understand this has been implemented by PR #674 

@dcalvoalonso should this issue be closed?

dcalvoalonso:
Sure! As always, thanks for your collaboration and feedback! :)

PR #673, opened 2018-09-27
Implements #672. Timezone is used to generate TimeInstant values

dcalvoalonso:
Implementation relies on [moment-timezone module] (https://momentjs.com/timezone/).

Appropriate unit tests have been added.

### DIFF
--- a/lib/services/ngsi/ngsiService.js
+++ b/lib/services/ngsi/ngsiService.js
@@ -36,6 +36,7 @@ var request = require('request'),
     utils = require('../northBound/restUtils'),
     config = require('../../commonConfig'),
     constants = require('../../constants'),
+    moment = require('moment-timezone'),
     logger = require('logops'),
     ngsiParser = require('./ngsiParser'),
     _ = require('underscore'),
@@ -252,11 +253,15 @@ function applyMiddlewares(middlewareCollection, entity, typeInformation, callbac
     }
 }
 
-function addTimestamp(payload) {
+function addTimestamp(payload, timezone) {
+    if (!timezone) {
+        timezone = 'UTC';
+    }
+
     var timestamp = {
             name: constants.TIMESTAMP_ATTRIBUTE,
             type: constants.TIMESTAMP_TYPE,
-            value: (new Date()).toISOString()
+            value: moment().tz(timezone).format('YYYY-MM-DD[T]HH:mm:ss.SSS') + 'Z'
         };
 
     function addMetadata(attribute) {
@@ -284,16 +289,18 @@ function addTimestamp(payload) {
 
     payload.contextElements[0].attributes.map(addMetadata);
     payload.contextElements[0].attributes.push(timestamp);
-
     return payload;
 }
 
-function addTimestampNgsi2(payload) {
+function addTimestampNgsi2(payload, timezone) {
 
-    function addTimestampEntity(entity) {
+    function addTimestampEntity(entity, timezone) {
+        if (!timezone) {
+            timezone = 'UTC';
+        }
         var timestamp = {
                 type: constants.TIMESTAMP_TYPE_NGSI2,
-                value: (new Date()).toISOString()
+                value: moment().tz(timezone).format('YYYY-MM-DD[T]HH:mm:ss.SSS') + 'Z'
             };
 
         function addMetadata(attribute) {
@@ -338,13 +345,13 @@ function addTimestampNgsi2(payload) {
     if (payload instanceof Array) {
         for (var i = 0; i < payload.length; i++) {
             if (!utils.isTimestampedNgsi2(payload[i])) {
-                payload[i] = addTimestampEntity(payload[i]);
+                payload[i] = addTimestampEntity(payload[i], timezone);
             }
         }
 
         return payload;
     } else {
-        return addTimestampEntity(payload);
+        return addTimestampEntity(payload, timezone);
     }
 
 }
@@ -459,7 +466,7 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
                         token);
 
                     if (config.getConfig().timestamp && !utils.isTimestampedNgsi2(result)) {
-                        result = addTimestampNgsi2(result);
+                        result = addTimestampNgsi2(result, typeInformation.timezone);
                     }
 
                     options.json = {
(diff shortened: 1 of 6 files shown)


lib/services/ngsi/ngsiService.js
```diff
@@ -36,6 +36,7 @@ var request = require('request'),
     utils = require('../northBound/restUtils'),
```
fgalan: CNR entry should be added

package.json
```diff
@@ -39,17 +39,18 @@
     "jison": "0.4.17",
     "logops": "1.0.0",
     "moment": "^2.19.4",
+    "moment-timezone": "^0.5.21",
```
fgalan: In order to avoid problems with "open" version dependencies, we prefer to use ~ instead of ^

(Some of pre-existing dependencies are yet using ^. We changed this in other repos but it seems this one was forgotten. Pre-existing dependencies will be fixed in a next PR).

test/unit/ngsiService/active-devices-test.js
```diff
@@ -247,6 +247,60 @@ describe('Active attributes test', function() {
         });
     });
 
+    describe('When the IoT Agent receives new information, the timestamp flag is on' +
+    'and timezone is defined', function() {
+        var modifiedValues;
+
+        beforeEach(function(done) {
+            var time = new Date(1438760101468);
```
fgalan: It would be useful a hint to which date this integer corresponds (in a commented line just above the line, for instance). The same for other possible cases in this PR.

test/unit/examples/contextRequests/updateContextTimestampTimezone.json
```diff
@@ -0,0 +1,41 @@
...
+      "attributes": [
+        {
+          "name": "state",
+          "type": "Boolean",
+          "value": "true",
+          "metadatas": [
+            {
+              "name": "TimeInstant",
+              "type": "ISO8601",
+              "value": "2015-08-05T00:35:01.468Z"
```
fgalan: As far as I understand, the functionality is about supporting timezones. However, in the JSON included in this PR "Z" is always used in mocked requests to CB.

Does it means that the IOTAgent is translation timezone and will always send datetimes in UTC to CB?

fgalan:
I have a general question, from the point of view of backward compatibility.

As far as I understand, currently (master) when we provision this way:

```
{
    ...
    "timezone": "Europe/Madrid" 
}
```

IOTAgent ignores it and interprets date in UTC. With the changes in this PR I understand this changes and from now on dates will be intepreted in the timezone corresponding Europe/Madrid (UTC+1)

So, could be some backward compatibility problem for system usings IOTAgents that are assuming the current behaviour (UTC) that would be suddenly see a change to UTC+1 when this PR lands into master?

lib/services/ngsi/ngsiService.js
dcalvoalonso: Fixed with 0f341d164437b4f2b9876ba8acf56d25afa5cefa

package.json
dcalvoalonso: Fixed with 0f341d164437b4f2b9876ba8acf56d25afa5cefa

test/unit/ngsiService/active-devices-test.js
dcalvoalonso: Fixed with 0f341d164437b4f2b9876ba8acf56d25afa5cefa

test/unit/examples/contextRequests/updateContextTimestampTimezone.json
dcalvoalonso: Sorry, to be honest, I did not know that the Z corresponds to a timezone. In fact, I thought that it was just part of the ISO 8601.
I have changed this and now formats are correctly formatted according to the standards and reflecting the timezone.

dcalvoalonso: Fixed with 0f341d164437b4f2b9876ba8acf56d25afa5cefa

fgalan: I'm not an expert in ISO8601, but I think that Z equals to UTC (a kind of "virtual timezone" where time is alwyas in UTC, no matter summer/winter time changes, etc.)

dcalvoalonso:
Your understanding of the current situation and about the implications of this PR is correct.

Nevertheless, I don't think that this is going to cause any backward compatibility. Users just will see that dates are appropriately formatted with respect to the provisioned timezone. If they want to go on using UTC, they just need to update the timezone for the device.

test/unit/examples/contextRequests/updateContextTimestampTimezone.json
dcalvoalonso: You're right, in fact, it means `Zulu time`: https://www.timeanddate.com/time/zones/z

:)

test/unit/examples/contextRequests/updateContextTimestamp.json
```diff
@@ -13,7 +13,7 @@
             {
               "name": "TimeInstant",
               "type": "ISO8601",
-              "value": "2015-08-05T07:35:01.468Z"
+              "value": "2015-08-05T07:35:01.468+00:00"
```
fgalan: I guess this is the case of a user that is not using timezone field at provision time (so it default to UTC).

Would it be possible to keep the current syntax for these cases? I mean, "2015-08-05T07:35:01.468Z" instead of "2015-08-05T07:35:01.468+00:00". I know that from a semantic point of view both are equivalent, but let's try to minimize the syntax change for current users not using timezone at all.

CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
 Add protocol in device provision if not provided (#652)
-Fix: timestamp is now included just in non empty (more than id and type) multienty entitites
\ No newline at end of file
+Fix: timestamp is now included just in non empty (more than id and type) multienty entitites
+Add: uses provisioned timezone to generate TimeInstant values (#672)
```
fgalan: Are timezones mentioned in documentation? I think the a description on how this functionality works would be a good addition (if not yet existing).

dcalvoalonso: Fixed with 07ae0f2a7ae383a35a28e9dcc00d9a827993d26b

package.json
```diff
@@ -39,17 +39,18 @@
     "jison": "0.4.17",
     "logops": "1.0.0",
     "moment": "^2.19.4",
+    "moment-timezone": "~0.5.21",
```
AlvaroVega: Can we fix moment-timezone version?

Issue #671, opened 2018-09-26
Payload for device provisioning


gabrieledeluca:
When I add a new device with _iot/devices_ POST request, I receive an error (WRONG_SYNTAX) when I use "forbidden" chars inside value of _object_id_ field (patterns are defined into _createDevice.json_ file). 
Why? Whereas I understand this pattern on other fields (such as name and type) linked with NorthBound world (and OCB forbidden characters), the _object_id_ should be linked to the SouthBound world so I expect to not have constraints (pattern) over _object_id_ value.

What do you think?

fgalan:
Northbound has a direct mapping to southbound with regards to attribute/measures. For instance, for active attributes, the measures IOTAgent gets in the southbound in an object_id are directly mapped to CB attributes in the northbound (in particular, to the attribute name associated to that object_id). So I think it makes sense that the restrictions in the northbound are applied also to northbound.

Or maybe I'm missing something... :)

gabrieledeluca:
In my opinion the OCB restrictions has not to be applied about the Southbound world: _object_id_ is the identification field of a Southbound thing and the library cannot know anything about Southbound restrictions (this is a "vertical" problem, so it has to be manage by a specific agent implementation).  
For example, if I have a Southbound stuff identified with this ID: "_thing=1_" I'll not be able to use this value as _object_id_ (without a workaround), why?

Instead, it is clear that there are some restrictions over _name_ field (as you have written above) because this field will be the entity attribute name in the Northbound.

Thanks in advance :-)

fgalan:
So your concern is related only with values of the measures/attributes (not the names). Is this correct? (just to undersatnd correctly the scope of the issue we are discussing here ;)

gabrieledeluca:
Yes, now there are restrictions over object_id that there should not be, so my concern is about don't have restrictions over the southbound world. This is a snippet from the _createDevice.json_ file: 

`"object_id": {
                  "description": "ID of the attribute in the device",
                  "type": "string",
                  "pattern": "^([^<>();'=\"]+)+$"
                },`

I don't understand why there is a pattern for an unknown (for the lib) device :-)

fgalan:
Let's try to clarify with an example. Let's suppose the following device has been provisioned:

```
{
  "devices": [ 
    { 
      "device_id": "DevID1", 
      "entity_name": "TheDevice1", 
      "entity_type": "DeviceType", 
      "attributes": [ 
[...]

PR #674, opened 2018-09-28
#671 issue

gabrieledeluca:
This solve Payload for device provisioning #671 issue using no pattern validation for object_id field in create e update device payload @fgalan :-)

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -47,7 +47,7 @@ module.exports = {
     DEFAULT_MONGODB_RETRIES: 5,
     DEFAULT_MONGODB_RETRY_TIME: 5,
 
-    ATTRIBUTE_DEFAULT: ' ',
+    ATTRIBUTE_DEFAULT: null,
 
     LOCATION_TYPE: 'geo:point',
     LOCATION_DEFAULT: '0, 0',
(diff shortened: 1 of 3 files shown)


gabrieledeluca:
committed also #675

fgalan:
With regards to #675 please find my feedback in the issue itself: https://github.com/telefonicaid/iotagent-node-lib/issues/675#issuecomment-426228935

With regards to #671 note that sometimes object_id is directly used for the attribute name. In particular, when no matching mapping (prevously defined in the provisioning operation used to created the devide) is found, then object_id is used. 

This PR relaxes the checking in two places (createDevice.json and updateDevice.json). Both cases apply only to the provision API (in which case it would be ok)? Is IOTAgent ensuring that for these object_ids without mapping no forbidden char is progresin in the name towards CB?

fgalan:
Any update on this PR @gabrieledeluca , please?

Just to know... :)

gabrieledeluca:
Hello @fgalan, nothing from my side: regarding the #671, I provided some examples in the OPC UA domain so I needed to relax the validation on the object_id field (the PR passed tests but I didn't write unit tests to check the potential forwarding of OCB forbidden characters). Instead regarding #675 I agree with you about backward compatibility although null value is semantically better.

jason-fox:
@gabrieledeluca - has this been superseded by. #920 ?

PR #702, opened 2018-11-07
Remove `##Index` from ToC since it is not required - fixes #648

### LINKED ISSUE #648: Maintain the ToC on Github whilst  hiding it on ReadtheDocs
Kind of related to the ToC discussion of #620 . I thought I'd raise an issue for discussion before trying anything out.

Q: Is the Table of Contents placed at the head of each markdown file:

1. Essential for navigating the documentation whilst on GitHub
2. Totally unnecessary and annoying whilst reading the documentation on ReadtheDocs
3. Both of the above

Personally I prefer viewing the documentation on RTD - I find the ToC pushes relevant information "below the fold" on each page - however I know it can't be removed since others navigate by using the GitHub Markdown.

I've been thinking that we could use CSS to style away the ToC on RTD by hiding the relevant `<ul>` as shown:

```css
h1:first-of-type+ul,
h2:first-of-type+ul{display: none}
```

This works fine and removes the contents of the ToC itself, but irritatingly would still leave the line `<h2>Contents</h2>` on screen - would it make sense to remove the `## Contents` line so both GitHub users and RTD users could be happy?

I mean do our users really need a sub-header to tell them that the list of links is a Table of Contents?

jason-fox:
* The ToC is suppressed with the ReadtheDocs CSS
* The ToC Header cannot be suppressed without CSS3

Therefore the title of the ToC should be removed to avoid confusion

### DIFF
--- a/doc/expressionLanguage.md
+++ b/doc/expressionLanguage.md
@@ -1,7 +1,5 @@
 # Measurement Transformation Expression Language
 
-## Index
-
 * [Overview](#overview)
 * [Measurement transformation](#measurement-transformation)
   + [Expression definition](#expression-definition)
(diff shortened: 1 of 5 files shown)


dcalvoalonso:
LGTM

PR #714, opened 2018-11-15
allow multientity with a multimeasure in a attribute with the same name

AlvaroVega:


dcalvoalonso:
LGTM! Great job @AlvaroVega !

PR #708, opened 2018-11-09
Checks ISO8601 format for timestamps

dcalvoalonso:
Implements #679.

Unit tests have been added to demonstrate the new functionality for NGSIv1 and NGSIv2.

PR #717, opened 2018-11-21
unit test with ISO8601 checking happy path

fgalan:
~~Maybe the implementation in PR https://github.com/telefonicaid/iotagent-node-lib/pull/708 had some bug. This PR (when finished) attemps to solve it.~~

Adding 4 unit test to check ISO8601 happy path

### DIFF
--- a/test/unit/ngsiService/active-devices-test.js
+++ b/test/unit/ngsiService/active-devices-test.js
@@ -423,6 +423,7 @@ describe('Active attributes test', function() {
         it('should not fail', function(done) {
             iotAgentLib.update('light1', 'Light', '', modifiedValues, function(error) {
                 should.not.exist(error);
+                contextBrokerMock.done();
                 done();
             });
         });
@@ -460,6 +461,7 @@ describe('Active attributes test', function() {
         it('should not fail', function(done) {
             iotAgentLib.update('light1', 'Light', '', modifiedValues, function(error) {
                 should.not.exist(error);
+                contextBrokerMock.done();
                 done();
             });
         });
--- a/test/unit/ngsiService/active-devices-test.js
+++ b/test/unit/ngsiService/active-devices-test.js
@@ -391,6 +391,80 @@ describe('Active attributes test', function() {
         });
     });
 
+    describe('When the IoTA gets a set of values with a TimeInstant which are in ISO8601 format', function() {
+        var modifiedValues;
+
+        beforeEach(function(done) {
+
+            modifiedValues = [
+                {
+                    name: 'state',
+                    type: 'Boolean',
+                    value: 'true'
+                },
+                {
+                    name: 'TimeInstant',
+                    type: 'ISO8601',
+                    value: '2022-22-22T22:22:22Z'
+                }
+            ];
+
+            nock.cleanAll();
+
+            iotAgentConfig.timestamp = true;
+            iotAgentLib.activate(iotAgentConfig, done);
+        });
+
+        afterEach(function(done) {
+            delete iotAgentConfig.timestamp;
+            done();
+        });
+
+        it('should not fail', function(done) {
+            iotAgentLib.update('light1', 'Light', '', modifiedValues, function(error) {
+                should.not.exist(error);
+                done();
+            });
+        });
+    });
+
+    describe('When the IoTA gets a set of values with a TimeInstant which are in ISO8601 format with ms', function() {
+        var modifiedValues;
+
+        beforeEach(function(done) {
+
+            modifiedValues = [
+                {
+                    name: 'state',
+                    type: 'Boolean',
+                    value: 'true'
+                },
+                {
+                    name: 'TimeInstant',
+                    type: 'ISO8601',
+                    value: '2022-22-22T22:22:22.222Z'
+                }
+            ];
+
+            nock.cleanAll();
+
+            iotAgentConfig.timestamp = true;
+            iotAgentLib.activate(iotAgentConfig, done);
+        });
+
+        afterEach(function(done) {
+            delete iotAgentConfig.timestamp;
+            done();
+        });
+
+        it('should not fail', function(done) {
+            iotAgentLib.update('light1', 'Light', '', modifiedValues, function(error) {
+                should.not.exist(error);
+                done();
+            });
+        });
+    });
+
     describe('When the IoTA gets a set of values with a TimeInstant, the timestamp flag is on' +
     'and timezone is defined', function() {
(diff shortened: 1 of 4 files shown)


fgalan:
Maybe `moment` is not working as expected? Test:

```
var moment = require('moment');

console.log("Test 2022-22-22T22:22:22.222Z :");
console.log(moment("2022-22-22T22:22:22.222Z", moment.ISO_8601).isValid());

console.log("Test 2022-22-22T22:22:22Z :");
console.log(moment("2022-22-22T22:22:22Z", moment.ISO_8601).isValid());
```

Result:

```
Test 2022-22-22T22:22:22.222Z :
false
Test 2022-22-22T22:22:22Z :
false
```

fgalan:
Month 22 was the cause of the problem. Very obvious, but the long sequene of digit 2 confused me :) Thus not actual bug at the end, but I think the PR is valuable (after adjusting it) to add 4 new unit tests to ensure the "happy path" of the ISO8601 checking functionality.

dcalvoalonso:
Do you need some help to extend the tests @fgalan ?

fgalan:
> Do you need some help to extend the tests @fgalan ?

I'm currently getting this problem in the tests

```
Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/fermin/src/iotagent-node-lib/test/unit/ngsiService/active-devices-test.js)
```

This kind of errors use to be due to trivial causes (not called callback, not called mock, etc.) but somtimes it's different to spot the cause. I haven't found the cause yet. If you could provide some hint, it would be great :) Thanks!

fgalan:
I think the problems are related with CB mock configuration. The tests in this PR were adapted from existing ones for the fail path (in which CB mock is not needed, as the request never progress to it). I'm on it.

test/unit/ngsiv2/ngsiService/active-devices-test.js
```diff
@@ -256,6 +256,82 @@ describe('Active attributes test', function() {
...
+
+        it('should not fail', function(done) {
+            iotAgentLib.update('light1', 'Light', '', modifiedValues, function(error) {
+                should.not.exist(error);
+                contextBrokerMock.done();
+                done();
+            });
+        });
+    });
+
```
fgalan: The above unit test happens to be the same as existing:

```
    describe('When the IoT Agent receives new information and the timestamp flag is on', function() {
        var modifiedValues;

        beforeEach(function(done) {
            var time = new Date(1438760101468); // 2015-08-05T07:35:01.468+00:00

            modifiedValues = [
                {
                    name: 'state',
                    type: 'boolean',
                    value: true
                },
                {
                    name: 'dimming',
                    type: 'number',
                    value: 87
                }
            ];

            timekeeper.freeze(time);

            nock.cleanAll();

            contextBrokerMock = nock('http://192.168.1.1:1026')
                .matchHeader('fiware-service', 'smartGondor')
                .post('/v2/entities/light1/attrs',
                    utils.readExampleFile('./test/unit/ngsiv2/examples/contextRequests/updateContextTimestamp.json'))
                .reply(204);

            iotAgentConfig.timestamp = true;
            iotAgentLib.activate(iotAgentConfig, done);
        });

        afterEach(function(done) {
            delete iotAgentConfig.timestamp;
            timekeeper.reset();

            done();
        });
```

So I understand is redundant and should be removod from this PR. @dcalvoalonso what do you think?

test/unit/ngsiService/active-devices-test.js
```diff
@@ -391,6 +391,82 @@ describe('Active attributes test', function() {
...
+
+        it('should not fail', function(done) {
+            iotAgentLib.update('light1', 'Light', '', modifiedValues, function(error) {
+                should.not.exist(error);
+                contextBrokerMock.done();
+                done();
+            });
+        });
+    });
+
```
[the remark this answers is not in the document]
fgalan: (While this is decided I have put the test in skip in 7c082ca3137d901d827bd20fda50894d73b9a275, so travis make work)

test/unit/ngsiv2/ngsiService/active-devices-test.js
fgalan: (While this is decided I have put the test in skip in 7c082ca3137d901d827bd20fda50894d73b9a275, so travis make work)

test/unit/ngsiService/active-devices-test.js
```diff
@@ -391,6 +391,99 @@ describe('Active attributes test', function() {
...
+            modifiedValues = [
+                {
+                    name: 'state',
+                    type: 'Boolean',
+                    value: 'true'
+                },
+                {
+                    name: 'TimeInstant',
+                    type: 'ISO8601',
+                    value: '2022-10-22T22:22:22Z'
```
fgalan: You may argue that this we already have a test for this, but the difference is that this one doesn't use miliseconds while the exsiting one use it. Thus, they are two different variants and I think it worth to keep both them.

(The same for the NGSIv2 tests)

fgalan:
Ready to review. Please @dcalvoalonso @AlvaroVega go ahead. Thx!

test/unit/ngsiService/active-devices-test.js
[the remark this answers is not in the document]
dcalvoalonso: Yes, for me it is redundant and must be removed. And sorry for not having helped before.

test/unit/ngsiv2/ngsiService/active-devices-test.js
dcalvoalonso: Yes, for me it is redundant and must be removed. And sorry for not having helped before.

test/unit/ngsiService/active-devices-test.js
dcalvoalonso: I agree! NTC

test/unit/ngsiService/active-devices-test.js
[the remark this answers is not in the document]
fgalan: Fixed in 956c980 (removing the test)

test/unit/ngsiService/active-devices-test.js
fgalan: Fixed in 9838184 (test removed)

test/unit/ngsiv2/ngsiService/active-devices-test.js
fgalan: Fixed in 9838184 (test removed)

Issue #699, opened 2018-11-07
Remove Node v4


fgalan:
(To be addresses in January 2019)

Node v4 will be removed by the end of 2018. In particular, this issue involves the following actions:

* Node v4 should be removed from .travis.yml
* engine version should be modified to be >=6 in packages.json
* logops library should be updated to 2.1.0 in packages.json

fgalan:
I'm also a big fan of linters :) However, when the linter goes beyong simple fixes like line length, indent, syntax cleanup, etc. and starts to involve changes in programming patterns (such as using `async await` instead of callbacks) things needs to be thougth with care.

Thus, I agree in doing it by stages, starting with the simple fixes and let deep changes for the end.

By the way, we are currently using `jshint` for linting. Should I understand that `eslint` is a better tool (I didn't know it)? Do you have some introductory reference about it or a comparison? Thanks!

jason-fox:
From https://www.sitepoint.com/comparison-javascript-linting-tools/

> ESLint is also the obvious choice if you want to use ES6 (or ES2015, as they seem to be calling it now). Of all the tools mentioned, it has the widest support for ES6 features.

We could introduce `Husky` with `JsLint` right away (to see that it performs correctly) and then update the linter to `Eslint` when Node v4 is finally dropped

fgalan:
Implemented by PR https://github.com/telefonicaid/iotagent-node-lib/pull/723

PR #719, opened 2018-11-22
Update package.json

AlvaroVega:


fgalan:
I think 2.1.0 doesn't work with Node v4 so while it is still supported I think we should use 1.0.8. The upgrade to 2.1.0 will be done in the context of issue https://github.com/telefonicaid/iotagent-node-lib/issues/699 (to be addressed by January 2019).

package.json
```diff
@@ -42,7 +42,7 @@
     "command-shell-lib": "1.0.0",
     "express": "~4.16.4",
     "jison": "0.4.18",
-    "logops": "1.0.0",
```
fgalan: Should be added to CNR

PR #720, opened 2018-12-03
Update commonConfig.js

AlvaroVega:


lib/commonConfig.js
```diff
@@ -114,6 +114,7 @@ function processEnvironmentVariables() {
     if (process.env.IOTA_CB_URL) {
```
fgalan: No CNR entry needed? Is related with something already include in CNR?

AlvaroVega: fixed in https://github.com/telefonicaid/iotagent-node-lib/pull/720/commits/88b51076fa24e5edcdcb443f69140235392b4c14

PR #697, opened 2018-11-06
Use trust and cbHost from deviceGroup in typeInformation

AlvaroVega:
related manager fix https://github.com/telefonicaid/iotagent-manager/pull/158

related issue https://github.com/telefonicaid/iotagent-node-lib/issues/685

Issue #685, opened 2018-10-23
createInitialEntityNgsi2 and other methods are not using the service group cbHost but the global one


chicco785:
this means that currently a different cbHost per deviceGroup is not properly supported

chicco785:
it looks to me that the problem is quite deep (the feature of cbHost for a specific device group was probably not properly tested), since device data used in the different communications are not "merged" anyhow with the group data.

chicco785:
I have already a partial fix, i need only to complete the tests (which
is more painful than solving the issue :P) hope to complete by the end
of this week

fgalan:
PR https://github.com/telefonicaid/iotagent-node-lib/pull/697 addresses this issues, at least partially.

@chicco785 what do you think? Is the fix in the same line than the one you proposed? Could the test you say are finalizing be a contribution to cover the fix?

fgalan:
After merging PR #697, could this issue be closed? Or is some piece of work pending and should keep opened?

@chicco785 what do you think?

PR #721, opened 2018-12-03
Implements missing aspects of use of group's cbhost

dcalvoalonso:
As it is explained in https://github.com/telefonicaid/iotagent-json/pull/353#issuecomment-442858364, there is an inconsistent use of the field `cbhost` of groups. Although thanks to https://github.com/telefonicaid/iotagent-node-lib/pull/697/files, it is used to create the initial entity in the CB using NGSIv2, to update entities and to send unregistration requests, it was not used to send the subscription request using NGSIv2, nor some of the previously mentioned operations using NSGSIv1.

This PR implements these missing aspects and modified the corresponding unit tests.

### DIFF
--- a/lib/services/devices/deviceService.js
+++ b/lib/services/devices/deviceService.js
@@ -318,8 +318,14 @@ function createInitialEntityNgsi2(deviceData, newDevice, callback) {
  * @param {Object} newDevice        Device object that will be stored in the database.
  */
 function createInitialEntityNgsi1(deviceData, newDevice, callback) {
+    var cbHost = config.getConfig().contextBroker.url;
+    if (deviceData.cbHost && deviceData.cbHost.indexOf('://') !== -1) {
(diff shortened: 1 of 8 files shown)


lib/services/devices/deviceService.js
```diff
@@ -318,8 +318,14 @@ function createInitialEntityNgsi2(deviceData, newDevice, callback) {
  * @param {Object} newDevice        Device object that will be stored in the database.
  */
 function createInitialEntityNgsi1(deviceData, newDevice, callback) {
```
fgalan: No CNR entry needed? Is related with something already include in CNR?

dcalvoalonso: This is related to `Fix: use trust and cbHost from deviceGroup (#685)`. I think that we don't need to add a new netry.

fgalan: Ok. NTC

Issue #691, opened 2018-10-30
Remove XML related code


dcalvoalonso:
As XML payloads are not used anymore by any IoT Agent nor NGSI, it would be good to remove xmldom dependency and all the code related to support this functionality.

This was agreed and discussed as part of https://github.com/telefonicaid/iotagent-node-lib/pull/682#pullrequestreview-169446079

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/724

PR #724, opened 2018-12-05
Removes XML related code

dcalvoalonso:
This PR implements #691.

- Unit tests are passed
- Linting is correct
- grep -Rin xml lib does not return any result
- grep -Rin xml test does not return any result


CHANGES_NEXT_RELEASE
```diff
@@ -28,3 +28,4 @@ Fix: mosquitto.conf.example file not found by iot/mosquitto Dockerfile (#554)
 Fix: isDomain is not used anymore for context availability registration (#701)
 Fix: checks ISO8601 timeinstants provided by devices (#679)
 Fix: corrects linting and includes npm run lint in travis
+Remove: xml related code and file (#691)
```
fgalan: Maybe better:

```
Remove: XML related code, dependencies and files (#691) 
```

dcalvoalonso: Fixed with 6046b6f0b7743a68e6cc3e4723596983eea9c4de

PR #726, opened 2018-12-11
Support authentication with Oauth2 using FIWARE Keyrock IDM (to master)

fgalan:
Follow up of PR #664, in order to merge the OAuth2 implementation into master. You can check it is a mirror copy (same +- diff lines).

@jason-fox and @chicco785 were pending on (now closed) PR #664 so it would be great if they can provide feedback or LGTM in this PR.

CC: @dcalvoalonso

### DIFF
--- a/doc/installationguide.md
+++ b/doc/installationguide.md
@@ -9,7 +9,7 @@ These are the parameters that can be configured in the global section:
 * **logLevel**: minimum log level to log. May take one of the following values: DEBUG, INFO, ERROR, FATAL. E.g.: 'DEBUG'.
 * **contextBroker**: connection data to the Context Broker (host and port). E.g.:
 
-```
+```javascript
         {
         host: '192.168.56.101',
         port: '1026'
@@ -18,7 +18,7 @@ These are the parameters that can be configured in the global section:
 
  * If you want to use NGSIv2 (only sending updates for active attributes):
 
-```
+```javascript
   {
     host: '192.168.56.101',
     port: '1026',
@@ -28,7 +28,7 @@ These are the parameters that can be configured in the global section:
 
 * **server**: configuration used to create the Context Server (port where the IoT Agent will be listening as a Context Provider and base root to prefix all the paths). The `port` attribute is required. If no `baseRoot` attribute is used, '/' is used by default. E.g.:
 
-```
+```javascript
         {
         baseRoot: '/',
         port: 4041
@@ -37,7 +37,7 @@ These are the parameters that can be configured in the global section:
 ```
 * **stats**: configure the periodic collection of statistics. Use `interval` in milliseconds to set the time between stats writings.
 
-```
+```javascript
     stats: {
         interval: 100
     }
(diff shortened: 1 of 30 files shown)


lib/services/common/securityServiceOAuth2.js
```diff
@@ -64,17 +71,27 @@ function auth(trust, callback) {
     };
 
     logger.debug(context, 'Authentication on the OAuth2 provider [%s] using trust [%s]', url, trust);
+    logger.debug(context, 'Using the following request:\n\n%s\n\n', JSON.stringify(options, null, 4));
```
jason-fox: The client secret is being pushed to the debug log in plain text. This is a security leak.

doc/installationguide.md
```diff
@@ -180,6 +180,7 @@ The following table shows the accepted environment variables, as well as the con
 | IOTA_AUTH_PORT            | authentication.port                 |
 | IOTA_AUTH_USER            | authentication.user                 |
 | IOTA_AUTH_PASSWORD        | authentication.password             |
+| IOTA_AUTH_PERMANENT_TOKEN | authentication.permanentToken       |
```
jason-fox: The `IOTA_AUTH_PERMANENT_TOKEN`, `IOTA_AUTH_PASSWORD` and `IOTA_AUTH_USER` are all sensitive items of information, and therefore should also be available as *Docker Secrets*

It shouldn't affect the PR here, but the Docker containers for the IoT Agents will need to handle loading passwords from an encrypted file.

An example `entrypoint.sh` which handles Docker secrets can be found within PostGres 

https://github.com/docker-library/postgres/blob/040949af1595f49f2242f6d1f9c42fb042b3eaed/11/docker-entrypoint.sh#L5-L25

doc/installationguide.md
```diff
@@ -85,6 +82,21 @@ These are the parameters that can be configured in the global section:
     }
 ```
 
+  Nevertheless, this kind of authentication relying on `refresh_token` grant type implies that when the acces_token expires, it is needed to request a new one from the IDM, causing some overhead in the communication with the Context Broker. To mitigate this issue, FIWARE KeyRock IDM  implements `permanent tokens` that can be retrieved using `scope=permanent`. With this approach, the IOTA does not need to interact with the IDM and directly include the `permanent token` in the header. In order to use this type of token, an additional parameter         `permanentToken` must be set to `true` in the `authentication` configuration. An environment variable `IOTA_AUTH_PERMANENT_TOKEN` can be also used for the same purpose. For instance:
+
+```
```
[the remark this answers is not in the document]
jason-fox: LGTM

lib/services/common/securityServiceOAuth2.js
jason-fox: LGTM

doc/installationguide.md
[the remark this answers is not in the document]
jason-fox: Thanks for that fix.

PR #744, opened 2019-01-18
Feature/add coverage

jason-fox:
Expose Code Coverage results to Coveralls. Runs `npm test:coverage` after the unit tests and then sends the results to [coveralls](https://coveralls.io)

the repo needs to be exposed on coveralls as well (obviously) - see https://coveralls.io/repos/new

- SHOULD requirement from TSC


package.json
```diff
@@ -55,6 +56,7 @@
     "underscore": "~1.9.1"
   },
   "devDependencies": {
+    "coveralls": "^3.0.2",
```
fgalan: We try to avoid `^` usage in packages.json dependencies (we have had bad experience with it in the past). Could you change it to `~`, please?

PR #526, opened 2016-11-23
Update uuid to version 3.0.0

marcbachmann:
## Version 3.0.0 of [uuid](https://github.com/kelektiv/node-uuid) just got published.

Hi there,

This week a new version of the [uuid](https://github.com/kelektiv/node-uuid) module got released. The old module which was available by the name [node-uuid](https://www.npmjs.com/package/node-uuid) got deprecated and will show error message upon every install of the module.

To get rid of those install errors I've created this **automated pull request**. I've run some checks against your code base to test whether you're using one of the [deprecated apis](https://github.com/kelektiv/node-uuid/commit/5ae7287fc935eb55ef39133e4be17ef623ca000e), but this isn't the case.

Please test the changes against your code. I didn't run any tests and therefore can't guarantee that it isn't breaking. You can also just close this pr if you don't want to update your module.

In case there's already another pr open to upgrade uuid, I'm sorry for the effort I'm causing.

PR #765, opened 2019-03-05
Finishes PR #526

dcalvoalonso:
As https://github.com/telefonicaid/iotagent-node-lib/pull/526 comes from `unknown repository` and seems to have no activity, I have created a new PR with the same changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "mongodb": "3.1.8",
     "mongoose": "5.3.6",
     "mu2": "~0.5.20",
-    "node-uuid": "~1.4.1",
+    "uuid": "^3.3.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "mongodb": "3.1.8",
     "mongoose": "5.3.6",
     "mu2": "~0.5.20",
-    "node-uuid": "~1.4.1",
+    "uuid": "^3.3.2",
```
fgalan: Let's use ~ instead of ^.

In addition, please add the change to CNR.

dcalvoalonso: Fixed with 2121f6bcf64e72ff8d52cd4394d2ece4705a18bd

PR #772, opened 2019-04-10
[WIP] Fix reconnection problems with single mongo db instances

aarranz:
This PR should fix reconnection with MongoDBs when using single MongoDBs instances (e.g. for basic or demo installations). In this scenario, if the connection to the MongoDB server is lost and the IoTAgent is not able to reconnect again in a short period of time, the IoTAgent node lib will never reconnect again to MongoDB leaving the IoTAgent in a zombi status. 

This can be easily reproduced by shutting down the MongoDB server while the IoTAgent is running, leaving the MongoDB server down enough time and starting it again.

This happens because the IoTAgent node libe is not listening to the `reconnectFailed` event. Mongoose will try to reconnect every `reconnectInterval` milliseconds for `reconnectTries` times before raising the `reconnectFailed` event (See [mongoose's documentation](https://mongoosejs.com/docs/4.x/docs/connections.html#options) for more details). In my opinion, the best option here is to make the IoTAgent process to quit as soon as this scenario is detected. Another option would be adding extra reconnection stuff, but I think it is not worth it due the fact that IoTAgent are usually executed using docker so you can configure a restart policy (in fact, the pm2 component currently installed on the docker images is doing the same and can be used outside the docker image).

This problem does not happen when using replica sets or sharding, as in those scenarios Mongoose will try to reconnect forever.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -7,4 +7,4 @@ Upgrade from node-uuid ~1.4.1 to uuid ~3.3.2.
 Add: IOTA_AUTH_URL, IOTA_AUTH_CLIENT_ID, IOTA_AUTH_CLIENT_SECRET and IOTA_AUTH_TOKEN_PATH env vars
 Fix: incomplete HTTPS support for NGSI subscriptions (#593)
 Add: support for authentication to NGSI subscription requests (#592) 
-
+Fix: process dies if reconnection to DB fails instead of remain in a zombie state (#772)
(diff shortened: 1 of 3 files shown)


fgalan:
Thanks for the contribution! Let me some time to check internally which option (either die or try to reconnect forever) we prefer.

fgalan:
After some internal discussion is ok the approach in the PR to die after reconnection fail. However, some minor comments to be included before merging the PR, please:

* Add the new error (MONGODB-004) to the list in doc/operations.md
* Uses exit code of 1, i.e. `process.exit(1)` I guess.
* Add a line to CHANGES_NEXT_RELEASE about the change. For instace: "Fix: process dies if reconnection to DB fails instead of remain in a zombie state"

Thanks!

aarranz:
Done, please review if the description for `MONGODB-004` is ok for you.

PR #776, opened 2019-04-22
"cgroups" literal in configuration group 

Jagatjot:


fgalan:
I see a lot of commit that include changes only in a single file ("Update xxx.js"). Why don't modify several related files in the same commit?

PR #784, opened 2019-04-30
Fix: removes warnings

dcalvoalonso:
This PR solves the short term task of https://github.com/telefonicaid/iotagent-node-lib/blob/master/docs/roadmap.md:
> Removal of warnings and deprecation warnings.


fgalan:
Thank so much for the contributions in this PR!

I wonder if something about deprecation warning removal should be mentioned in CHANGES_NEXT_RELEASE. Maybe the change is minor... but for the sake of compleness maybe is a good idea.

PR #792, opened 2019-06-11
Updates installation guide for NGSIv2

### LINKED ISSUE #791: ngsi v2 configuration
looking at the iot agent lib, it seems that configuration should specify the protocol used as follow:
```
        contextBroker: {
            host: '192.168.1.1',
            port: '1026',
            ngsiVersion: 'v2'
        }
```

This is not documented in any agent neither in the lib, afaik.

dcalvoalonso:
Fixes #791

### DIFF
--- a/doc/installationguide.md
+++ b/doc/installationguide.md
@@ -16,7 +16,7 @@ These are the parameters that can be configured in the global section:
         }
 ```
 
- * If you want to use NGSIv2 (only sending updates for active attributes):
+ * If you want to use NGSIv2 (sending updates for active attributes and NGISv2-based registration):
 
 ```javascript
   {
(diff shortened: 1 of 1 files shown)


doc/installationguide.md
```diff
@@ -16,7 +16,7 @@ These are the parameters that can be configured in the global section:
         }
 ```
 
- * If you want to use NGSIv2 (only sending updates for active attributes):
+ * If you want to use NGSIv2 (sending updates for active attributes and NGISv2-based registration):
```
fgalan: Maybe the "(...)" should be removed at all. I think we are near to finalize the last part of NGSIv2 (support in forwarded queries) and if we leave the "(...)" note it could become obsolete soon and we may forget removing it.

dcalvoalonso: Fixed with ae87e84e298bbf60882f425713c41f1795209bb4

PR #822, opened 2019-10-30
not to use expiration in NGSIv2 registrations and expirations + throttling removal

fgalan:
Registraionts/subscriptions should be non-expirabale (as the IOTA deals with its removal when the device is removed). It doesn't make sense to keep using expiration in the case of NGSIv2 (as the API allows that, different from NGSIv1, which mandates to use duration).

In additon, throttling can be problematic in command & lazy attirbutes scenarios. Removed (in both NGSIv1 and NGSIv2)

- [x] Code
- [x] Tests
- [x] Doc review (look for "expiration", "duration" and "throttling" occurrences)


doc/installationguide.md
```diff
@@ -142,7 +142,7 @@ are 5 and 5 respectively). E.g.:
...
 * **subservice**: default subservice for the IoT Agent. If a device is being registered, and no subservice information comes with the device data, and no subservice information is configured for the given type, the default IoT agent subservice will be used instead. E.g.: '/gardens'.
 * **providerUrl**: URL to send in the Context Provider registration requests. Should represent the external IP of the deployed IoT Agent (the IP where the Context Broker will redirect the NGSI requests). E.g.: 'http://192.168.56.1:4041'.
-* **deviceRegistrationDuration**: duration of the registrations as Context Providers, in [ISO 8601](http://en.wikipedia.org/wiki/ISO_8601) standard format. E.g.: 'P1M'.
+* **deviceRegistrationDuration**: duration of the registrations as Context Providers and the subscriptions done by bidirectional plugin, in [ISO 8601](http://en.wikipedia.org/wiki/ISO_8601) standard format. E.g.: 'P1M'. Only used if agent is configured in NGSIv1 mode. In NGSIv2 mode the setting is ignored (registrations and subscriptions are non-expirable and the IOTA manages its removal explicitely).
```
fgalan: Note there is no modification regarding throttling in documentation... due to throttling was never documented :)

NTC

PR #836, opened 2020-01-09
Add NGSIv2 metadata support to device provisioned attributes

jason-fox:
I want to be able to add metadata when provisioning an IoT Agent, so that the units of measurement can be placed into the resultant entity.

e.g.:

```json
{
     "entity_type": "Lamp",
     "resource":    "/iot/d",
..etc
     "commands": [ 
        {"name": "on","type": "command"},
        {"name": "off","type": "command"}
     ],
     "attributes": [
        {"object_id": "s", "name": "state", "type":"Text"},
        {"object_id": "l", "name": "luminosity", "type":"Integer",
          "metadata":{
              "unitCode":{"type": "Text", "value" :"CAL"}
          }
        }
     ],
     "static_attributes": [
          {"name": "category", "type":"Text", "value": ["actuator","sensor"]},
          {"name": "controlledProperty", "type": "Text", "value": ["light"],
            "metadata":{
              "includes":{"type": "Text", "value" :["state", "luminosity"]},
              "alias":{"type": "Text", "value" :"lamp"}
            }
          },
     ]
   }
```

Currently `metadata` placed into the provisioning request is **valid**, but the `metadata` is not **processed** when exchanging requests with the Context Broker.  Only `metadata` coming directly from a **Measure** is  exchanged.

Furthermore, typical NGSI v2 `metadata` is in the form of key-value pairs - a.k.a _properties-of-properties_. These values should also be processed properly along with NGSI v1 style metadata-as-an- array.

This PR does the following:

- Ensure static attribute and active attributes can hold metadata
- Process metadata objects as well as arrays.
- Ensures that Generated Timestamps do not obliterate other metadata elements.
- Create Tests

### DIFF
--- a/lib/plugins/pluginUtils.js
+++ b/lib/plugins/pluginUtils.js
@@ -106,7 +106,9 @@ function createProcessAttribute(fn, attributeType) {
         }
 
         if (config.checkNgsi2()) {
-
+            // This code is backwards compatible to process metadata in the older NGSIv1-style (array) 
+            // as well as supporting the newer NGSIv2-style (object). The redundant Array Check can be 
+            // therefore be removed if/when NGSIv1 support is removed from the library.
             if (Array.isArray (attribute.metadata)){
                 attribute.metadata = attribute.metadata.map(createProcessAttribute(fn, attributeType));
             } else if (attribute.metadata) {
--- a/lib/plugins/pluginUtils.js
+++ b/lib/plugins/pluginUtils.js
@@ -106,8 +106,14 @@ function createProcessAttribute(fn, attributeType) {
         }
 
         if (config.checkNgsi2()) {
-            if (attribute.metadata) {
+
+            if (Array.isArray (attribute.metadata)){
                 attribute.metadata = attribute.metadata.map(createProcessAttribute(fn, attributeType));
+            } else if (attribute.metadata) {
+                Object.keys(attribute.metadata).forEach(function(key) {
+                    var entry = attribute.metadata[key];
+                    entry.value = fn(entry.value);
+                });
             }
         } else {
             if (attribute.metadatas) {
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,2 +1,3 @@
+Add NGSIv2 metadata support to device provisioned attributes
 Fix: Error message when sending measures with unknown/undefined attribute
 Add Null check within executeWithSecurity() to avoid crash 
--- a/lib/plugins/timestampProcessPlugin.js
+++ b/lib/plugins/timestampProcessPlugin.js
@@ -45,7 +45,7 @@ function updatePluginNgsi2(entity, entityType, callback) {
     function insertMetadata(element) {
 
         if (element.name !== constants.TIMESTAMP_ATTRIBUTE) {
-            var metadata = {};
+            var metadata = element.metadata || {};
             metadata[constants.TIMESTAMP_ATTRIBUTE] = {
                 type: constants.TIMESTAMP_TYPE_NGSI2,
                 value: timestamp.value
--- a/lib/services/commands/commandRegistryMongoDB.js
+++ b/lib/services/commands/commandRegistryMongoDB.js
@@ -56,8 +56,7 @@ function findCommand(service, subservice, deviceId, name, callback) {
             callback(null, data);
         } else {
             logger.debug(context, 'Command for DeviceID [%j] with name [%j] not found', deviceId, name);
-
-            callback(new errors.CommandNotFound());
+            callback(new errors.CommandNotFound(name));
         }
     });
 }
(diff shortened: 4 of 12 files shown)


fgalan:
Just to confirm... this is a kind of "static metadata" feature? I mean, IOTA will send exactly the metadata-value included in the provisioning request for the device (vs. dynamically set the metadata value somehow from the measure sent by the device).

With regards to

> Ensures that Generated Timestamps do not obliterate other metadata elements.

Does it means that the timestamp metadata generated by IOTA will override timestamp metadata provided in the provisioning operation? Or the other way arround?

manucarrace:
Seems ok for me, specially for static attributes. On the other hand, ss far as I know, it is not possible to send metadata to IoTAgent in a measure so, metadata in active attributes would be a kind of static information set at provision time.

Please, could you clarify if metadata for an active attribute would be sent only at provision time or with every measure?

jason-fox:
> Just to confirm... this is a kind of "static metadata" feature? 

Yes, I would envisage you would only add static metadata at provision time such as the `metadata.unitCode` for the units of an attribute, or some static info needed to help manipulate the data such as `metadata.accuracy`.  It would be very silly to have a dynamic value here such as `metadata.TimeInstant`. I'd also like to use it for some NGSI v2 => NGSI-LD conversion work I'm looking at, but the static data to be provided will be down to the developer.

> Does it means that the timestamp metadata generated by IOTA will override timestamp metadata

It **used** to be the case that the **only** `metadata` was `metadata.TimeInstant` and this was created by stating `metadata = {}`. The code has been amend to start with `metadata = static metadata` and then set the `TimeInstant` value so that my static metadata isn't removed if Timestamps are being used, The `metadata.TimeInstant` will be the dynamic timestamp of course.

> as far as I know, it is not possible to send metadata to IoTAgent in a measure...

If you look carefully at the amendments to `active-devices-test.js`, you will see that there is/was test which makes the assumption that an IoT Agent is capable of passing metadata along with a measure. I don't think any of them do so at the moment, though someone has [requested it](https://stackoverflow.com/questions/59339190/fiware-metadata-in-iotagent). For some reason the code made the assumption that metadata within an NGSI v2  attribute was an **array**, which I believe is incorrect (or a hangover from NGSI v1 `metadatas`) - I've added an Array check in `pluginUtils.js` to maintain the backwards functionality whilst also being able to process metadata Objects properly.

> metadata for an active attribute ...  only at provision time or with every measure?

If metadata has been defined in an attribute, it will be passed whenever the measure is updated.

#### No metadata

```json
"temperature": {
         "value" : "20",
          "type":"Text"
}
```

#### Static metadata

```json
"temperature": {
         "value" : "20",
          "type":"Text",
          "metadata":{
              "unitCode":{"type": "Text", "value" :"CEL"}
          }
}
```

#### Static metadata plus timestamp

```json
"temperature": {
         "value" : "20",
          "type":"Text",
          "metadata":{
             "unitCode":{"type": "Text", "value" :"CEL"},
             "TimeInstant":{ "type": "DateTime", "value": "2015-08-05T07:35:01.468Z" }
      }
}
```

lib/plugins/pluginUtils.js
```diff
@@ -106,8 +106,14 @@ function createProcessAttribute(fn, attributeType) {
...
         if (config.checkNgsi2()) {
-            if (attribute.metadata) {
+
+            if (Array.isArray (attribute.metadata)){
                 attribute.metadata = attribute.metadata.map(createProcessAttribute(fn, attributeType));
+            } else if (attribute.metadata) {
+                Object.keys(attribute.metadata).forEach(function(key) {
+                    var entry = attribute.metadata[key];
+                    entry.value = fn(entry.value);
+                });
```
fgalan: I guess this modification is related with:

> For some reason the code made the assumption that metadata within an NGSI v2 attribute was an **array**, which I believe is incorrect (or a hangover from NGSI v1 `metadatas`) - I've added an Array check in `pluginUtils.js` to maintain the backwards functionality whilst also being able to process metadata Objects properly.

I think is the later (a hanover from NGSIv1 metadatas).

Given you have prepared the code to cover both cases (great!), maybe it a good idea to include a commenta about it (something like: "This code is prepared to process metadata in NGSIv1-style (array) and NGSIv2-style (object)"). This way it would be easy to remove the NGSIv1 part if we remove at the end all NGSIv1 stuff from agents code.

fgalan:
@jason-fox thank you very much for so detailed explanation.

Apart from a little comment I have provided in the diff tag, some general ones:

* An entry about the changes introduced in this PR should be added to CHANGES_NEXT_RELEASE
* What about documentation? I understand the new capabilities of the provisioning API to set "static metadata" should be described (which exact .md file or files to touch should be analyzed)

Thanks!

lib/plugins/pluginUtils.js
jason-fox: Clarification added. Fixed a2d139

jason-fox:
@fgalan - I've added the comment and updated the CNR. I'll raise a separate PR for documentation as I#d prefer to keep this PR small and atomic so it is easier to merge.

PR #838, opened 2020-01-16
Correct indentations within the documentation code

jason-fox:
- Remove extra whitespace to  ensure brackets line up.
- Standardize code snippets to use a common convention (e.g. single quote)


doc/finalResult.js
```diff
@@ -22,11 +22,11 @@ function parseUl(data, device) {
...
                 type: findType(pair[0])
             };
-        
+
         return attribute;
     }
-    
-    return data.split("#").map(createAttribute);
+
+    return data.split('#').map(createAttribute);
```
fgalan: I understand that single quote convention is the one we mainly use along .js code except for the few cases fixed in this PR.

Is my understanding correct?

fgalan: NTC

PR #843, opened 2020-02-10
Add GeoJSON and DateTime, unitCode and observedAt NGSI-LD support

jason-fox:
This PR adds GeoJSON and DateTime,`unitCode` and `observedAt`  NGSI-LD support 

Parsing now supports native parsing of JSON primitives plus the following: 

### Temporal Properties
-   `DateTime`  - always expressed in UTC, using ISO 8601 
-  `Date` - always expressed in UTC, using ISO 8601 
- `Time` - always expressed in UTC, using ISO 8601 

### Geospatial Properties
-  `Point` - represent as `coordinates` as mandated by IETF RFC 7946 
-  `LineString` - represent as `coordinates` as mandated by IETF RFC 7946 
- `Polygon` - represent as `coordinates` as mandated by IETF RFC 7946 
- `MultiPoint` - represent as `coordinates` as mandated by IETF RFC 7946 
- `MultiLineString` - represent as `coordinates` as mandated by IETF RFC 7946 
- `MultiPolygon` - represent as `coordinates` as mandated by IETF RFC 7946 

The NGSI v2 `TimeInstant` element has been mapped onto the NGSI-LD `observedAt` property

The NGSI v2 `metadata.unitCode` attribute has been mapped onto the NGSI-LD `unitCode` property

The set of tests expectation have been amended where necessary. Additional tests added.

PR #835, opened 2019-12-12
updated src and UT files for implementation of "selectively ignore some measure"

Madhu-NEC:
PR is with correspond to [iotagent-json#416](https://github.com/telefonicaid/iotagent-json/issues/416).

PR #847, opened 2020-02-25
Add NGSI-LD multi-measures support

jason-fox:
- split NGSI-v1, NGSI-v2 and NGSI-LD into separate files
- Move common REST functions into restUtils.js
- ensure that ids are URNS (adding urn:ngsi-ld: where necessary)
- Use Batch Upsert endpoint throughout ( `/ngsi-ld/v1/entityOperations/upsert/`) instead of PATCH update
- Update test expectations
- Split NGSI v2 and NGSI-LD registrations


fgalan:
PR #843 has been merged and base branch changed to feature/842_ngsi_ld in this PR.

So I understand you should upgrade jason-fox:feature/ngsi-ld-multimeasure with telefonicaid:feature/842_ngsi_ld

lib/services/devices/registrationUtils.js
```diff
@@ -317,15 +314,18 @@ function sendUnregistrationsNgsiLD(deviceData, callback) {
...
-        return deviceService.executeWithSecurity(
+        return utils.executeWithSecurity(
             options,
             deviceData,
-            createRegistrationHandlerNgsi2(true, deviceData, callback));
+            createRegistrationHandlerNgsi2(true, deviceData, callback)
+        );
     }
 
+    //console.error(JSON.stringify(options.json, null, 4));
```
fgalan: Leftover comment?

lib/services/common/genericMiddleware.js
```diff
@@ -97,7 +97,7 @@ function getLogLevel(req, res, next) {
 /**
  * Ensures the request type is one of the supported ones.
  */
-function ensureType(req, res, next) {
+function ensureType(req, res, next) {
```
fgalan:
```suggestion
function ensureType(req, res, next) {
```

fgalan:
I think the PR is ok (except for a couple of very minor comments easy to fix)

However, regarding:

> split NGSI-v1, NGSI-v2 and NGSI-LD into separate files

I like the refactor to move separate parts into -NGSI-LD.js -NGSI-v1.js and- NGSI-v2.js files in different parts. However, I wonder about the usual question: which impact could have this change into other existing PRs? Could you analyze it a little bit and provide feedback, pls?

lib/services/devices/registrationUtils.js
jason-fox: Yes, but - the multiple instances of `console.error`  (commented out) are being used to help generate the responses for the tests.

When the final PR is processed I'll remove them all.

lib/services/common/genericMiddleware.js
jason-fox: Not sure how that whitespace occured.

jason-fox:
> split NGSI-v1, NGSI-v2 and NGSI-LD into separate files

This is very simple - All of the _split_ `/devices`, `/ngsi` and `/northBound` files are to do with **NGSI Interactions only** - these are all defined by the relevant specifications, so there is zero likelihood that another PR will be amending them.  It will have **NO* impact on the other PRs. The split has not changed the functionality of the underlying code (since none of the v1 or v2 tests has been modified) - its merely cut-paste rearrangement.

Going forward,  reducing the file size means that it is even less likely that a clash will occur - besides which  it is practically impossible to read a single 2000+ line file with interleaving implementations of three  near duplicates of each function with only fractional differences between them. [sigh]

jason-fox:
Note that things like https://github.com/telefonicaid/iotagent-node-lib/pull/835  can touch files like `deviceService.js`, but this is **common** functionality which applies to  v1 v2 and LD and therefore the relevant functions remain unchanged in the root file.

lib/services/devices/registrationUtils.js
fgalan: Noted in the accumulative PR - https://github.com/telefonicaid/iotagent-node-lib/pull/849#issuecomment-594666173

NTC with regards this PR

lib/services/common/genericMiddleware.js
fgalan: Fixed in cbf6cef

PR #841, opened 2020-02-07
Basic NGSI-LD active measures support

jason-fox:
This PR offers NGSI-LD support for **active measures only** under the following configuration.

```javascript
contextBroker: {
            url: 'https://192.168.1.1:1026',
            ngsiVersion: 'ld',
            jsonLdContext: 'http://context.json-ld'
        },
```

The existing NGSI-v2 support remains as default under `ngsiVersion: 'v2' `

The NGSI-LD support matches the decision trees  for NGSI v1/v2 - where necessary a third option has been added and the necessary functionality **duplicated**.

The set of tests from NGSI v2 have been duplicated and the expectations amended where necessary.

Commands, Subscriptions and Lazy Attribute Support have not been added - disabled test stubs have been created to allow them to be tested once the code base has been updated.

PR #831, opened 2019-11-30
Feature/eslint

jason-fox:
- Replace jshint with eslint
- Add prettier code formatting
- Add husky and lint-staged

Incorporates and supercedes https://github.com/telefonicaid/iotagent-node-lib/pull/753

Related: https://github.com/telefonicaid/iotagent-node-lib/issues/830

PR #848, opened 2020-02-27
Add NGSI-LD basic Command support

jason-fox:
This PR adds basic Command and Lazy Attribute support

- Split out NGSI-v1, v2 and LD handlers into separate files.
- Add support for PATCH incoming commands.
- Add support for standard NGSI-LD Headers
- Amend tests ( replace `/v2/op/update/`  with NGSI-LD PATCH)
- Amend tests ( replace `/v2/op/query/`  with NGSI-LD GET)
- Amend test expectations to NGSI-LD format

PR #832, opened 2019-11-30
Fix #812 Catch missing file failure and continue

jason-fox:
The *failure-to-load-a-file* error is occuring before any test is running and hence the number of failed tests=0.  Mocha completes, since no tests have failed this is interpreted as **success**. 

By amending the `readExampleFile()` utility function to catch all errors, attempting to load a non-existent file does not error - it just gives the wrong expectation and therefore should lead the test to fail.

### Rationale

Running

```console
mocha --recursive 'test/**/*.js' --reporter spec --timeout 3000 --ui bdd --exit
 echo $?
```

Returns `0` so the test suite runs successfully even though the following error occurs:

```text
{ Error: ENOENT: no such file or directory, open './test/unit/examples/deviceProvisioningRequests/provisionAutoprovision.json'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.readFileSync (fs.js:551:33)
    at Object.readExampleFile (./iotagent-node-lib/test/tools/utils.js:27:21)
    at Suite.<anonymous> (./iotagent-node-lib/test/unit/ngsiv2/provisioning/device-provisioning-api_test.js:315:25)
    at Object.create (/usr/local/lib/node_modules/mocha/lib/interfaces/common.js:112:19)
    at context.describe.context.context (/usr/local/lib/node_modules/mocha/lib/interfaces/bdd.js:40:27)
    at Suite.<anonymous> (./iotagent-node-lib/test/unit/ngsiv2/provisioning/device-provisioning-api_test.js:311:5)
    at Object.create (/usr/local/lib/node_modules/mocha/lib/interfaces/common.js:112:19)
    at context.describe.context.context (/usr/local/lib/node_modules/mocha/lib/interfaces/bdd.js:40:27)
    at Object.<anonymous> (./iotagent-node-lib/test/unit/ngsiv2/provisioning/device-provisioning-api_test.js:52:1)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at /usr/local/lib/node_modules/mocha/lib/mocha.js:250:27
    at Array.forEach (<anonymous>)
    at Mocha.loadFiles (/usr/local/lib/node_modules/mocha/lib/mocha.js:247:14)
[...]

PR #850, opened 2020-03-04
Update NGSI support to use ES6 syntax.

jason-fox:
Similar to https://github.com/telefonicaid/iotagent-node-lib/pull/841#issuecomment-584621453

This   is a **limited** PR updates the files within  `/services/devices/`, `services/ngsi/` and `services\northBound` to properly use `let` and `const`. When processed it would reduce the scope of #831

It only contains automated code updates at the moment. The PR could also be amended to correctly rename the `northBound` subdirectory to `northPort` see: #645

The assumption here is that since none of the other outstanding PRs are amending the NGSI interactions, this would be an opportunity to modernize at least part of the code base.


fgalan:
Once PR #848 has been merged, I have change the base of this PR to feature/842_ngsi_ld

Please upgrade jason-fox:feature-ngsi-ld-es6 with telefonicaid:feature/842_ngsi_ld so this PR gets ready to merge (note a couple of files are conflicting).

jason-fox:
Merged, updated and removed `console.error()` debug - see https://github.com/telefonicaid/iotagent-node-lib/pull/847#discussion_r387775415

test/tools/utils.js
```diff
@@ -37,11 +35,6 @@ function readExampleFile(name, raw) {
         console.error(JSON.stringify(e));
```
fgalan: Should this one also be removed? Or is pre-existing to your work on NGSI-LD?

(Maybe a grep for `console.` in all the code base could help to find additional cases)

jason-fox: Nope - that is a deliberate console output - see the previous line. It was part of the output of #832

PR #852, opened 2020-03-18
fix the mongodb authentication

SamuelTJackson:
With this pr the mongodb authentication is fixed. Here I have incorporated the suggestions of @srrspace mentioned here https://github.com/telefonicaid/iotagent-node-lib/commit/13f67ead5327ebdcd97a7b7858ac64a361ad9655 . Username, Password, AuthSource and ReplicaSet are no longer written in the connection string but in the options. The option parameter are looked up here http://mongodb.github.io/node-mongodb-native/2.2/api/MongoClient.html#connect .

I tested it with a mongodb with replicaset and authentication and without replicaset and authentication.

### DIFF
--- a/lib/model/dbConn.js
+++ b/lib/model/dbConn.js
@@ -51,19 +51,15 @@ function loadModels() {
  *
  * @this Reference to the dbConn module itself.
  */
-function init(host, db, port, username, password, options, callback) {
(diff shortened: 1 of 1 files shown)


srrspace:
It's OK for me, I have tested these changes and now mongodb authentication works fine with iotagent-json.

PR #863, opened 2020-05-06
fix jexl transformation expression when is  a JSON based value

AlvaroVega:
bug in previos PR https://github.com/telefonicaid/iotagent-node-lib/pull/861
to avoid json expression will be transformed to [object Object]

```
{
            "name" : "mylocation",
            "type" : "geo:json",
            "expression" : "{type: \"Point\", coordinates: [lat,lon]}"
        }
```

No CNR update is needed since is reopen a previous PR

### DIFF
--- a/lib/plugins/jexlParser.js
+++ b/lib/plugins/jexlParser.js
@@ -133,12 +133,12 @@ function expressionApplier(context, typeInformation) {
         else if (attribute.type === 'None') {
             newAttribute.value = null;
         }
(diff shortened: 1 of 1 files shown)


chicco785:
i think it may be good to add this in the test and examples

fgalan:
> i think it may be good to add this in the test and examples

Sound goods. Maybe you could help providing such contribution? We can focus our resources in fix some flaws of your original implementation, but for things like those your contribution would be very welcomed!

PR #849, opened 2020-03-02
[e2e testing] NGSI-LD

fgalan:
This PR accumulates the NGSI-LD contribution, previous to its merging in master. It grows with the merging of other PRs into feature/842_ngsi_ld and we can check how travis CI evolve merge after merge. At the end, an e2e CI pass will be done on this PR branch.

PRs included so far:

* PR https://github.com/telefonicaid/iotagent-node-lib/pull/841: Basic NGSI-LD active measures support
* PR https://github.com/telefonicaid/iotagent-node-lib/pull/843: Add GeoJSON and DateTime, unitCode and observedAt NGSI-LD support
* PR https://github.com/telefonicaid/iotagent-node-lib/pull/847: Add NGSI-LD multi-measures support
* PR #848: Add NGSI-LD basic Command support
* PR #850: Update NGSI support to use ES6 syntax

CC: @jason-fox

PR #882, opened 2020-07-17
Make missing params public

jason-fox:
Fix related to review comment: https://github.com/telefonicaid/iotagent-node-lib/pull/849#discussion_r446045335

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -73,6 +73,8 @@ module.exports = {
     MONGO_ALARM: 'MONGO-ALARM',
     ORION_ALARM: 'ORION-ALARM',
     IOTAM_ALARM: 'IOTAM-ALARM',
+    ATTRIBUTE_DEFAULT,
+    DATETIME_DEFAULT,
 
     getInitialValueForType
 };


lib/constants.js
```diff
@@ -73,6 +73,8 @@ module.exports = {
     MONGO_ALARM: 'MONGO-ALARM',
     ORION_ALARM: 'ORION-ALARM',
     IOTAM_ALARM: 'IOTAM-ALARM',
+    ATTRIBUTE_DEFAULT,
+    DATETIME_DEFAULT,
```
fgalan, line 77: Should these constants have an actual value (as the others above)?

fgalan: Maybe it's done in L29 and L30 ?

```
const DATETIME_DEFAULT = '1970-01-01T00:00:00.000Z';
const ATTRIBUTE_DEFAULT = ' ';
```

jason-fox: Correct - standard scope inheritance.

fgalan: NTC

PR #885, opened 2020-07-31
Security improvement: Move secret support within Node

jason-fox:
Currently the Docker Secret and Password processing lies within an entrypoint bash file e.g.: 

https://github.com/telefonicaid/iotagent-ul/blob/master/docker/entrypoint.sh

From a security perspective this isn't ideal, as the dockerization will be Node.js + bash script combination. This means the application can never be run *Distroless* as it relies on bash, and therefore offers a wider attack profile.

https://medium.com/@dwdraju/distroless-is-for-security-if-not-for-size-6eac789f695f

This PR moves as much of the bash scripting as possible inside the Node app, which will allow for the creation of smaller, securer *Distroless* Docker images.

A sample *Distroless* build can be found here: https://github.com/FIWARE/tutorials.Step-by-Step/blob/master/docker/Dockerfile 

```Dockerfile
# Create an anonymous user
RUN sed -i -r "/^(root|nobody)/!d" /etc/passwd /etc/shadow /etc/group \
    && sed -i -r 's#^(.*):[^:]*$#\1:/sbin/nologin#' /etc/passwd


# Remove the distro as part of a two stage build.
FROM gcr.io/distroless/nodejs:${NODE_MAJOR_VERSION}
COPY --from=build-env /usr/src/app /usr/src/app
USER nobody
WORKDIR /usr/src/app

# Ports used by application
EXPOSE ${WEB_APP_PORT:-3000} ${DUMMY_DEVICES_PORT:-3001}
# Start the application
CMD ["./bin/www"]

HEALTHCHECK NONE
```

This style of Dockerization could be ported over to the IoT Agents themselves and the bash script removed.

### DIFF
--- a/lib/commonConfig.js
+++ b/lib/commonConfig.js
@@ -179,18 +179,18 @@ function processEnvironmentVariables() {
             'IOTA_MONGO_PASSWORD'
         ];
 
+    // Substitute Docker Secret Variables where set.
+    protectedVariables.forEach((key) => {
+        getSecretData(key);
+    });
     environmentVariables.forEach((key) => {
         let value = process.env[key];
         if (value) {
-            if (key.endsWith('USER') || key.endsWith('PASSWORD')) {
+            if (key.endsWith('USER') || key.endsWith('PASSWORD') || key.endsWith('CLIENT_ID') || key.endsWith('CLIENT_SECRET')) {
                 value = '********';
             }
             logger.info('Setting %s to environment value: %s', key, value);
         }
-    });
-    // Substitute Docker Secret Variables where set.
-    protectedVariables.forEach((key) => {
-        getSecretData(key);
     });
 
     // Context Broker Configuration (ensuring the configuration sub-object exists before start using it)
--- a/lib/commonConfig.js
+++ b/lib/commonConfig.js
@@ -186,7 +186,10 @@ function processEnvironmentVariables() {
     environmentVariables.forEach((key) => {
         let value = process.env[key];
         if (value) {
-            if (key.endsWith('USER') || key.endsWith('PASSWORD') || key.endsWith('CLIENT_ID') || key.endsWith('CLIENT_SECRET')) {
+            if (key.endsWith('USER') || 
+                key.endsWith('PASSWORD') || 
+                key.endsWith('CLIENT_ID') || 
+                key.endsWith('CLIENT_SECRET')) {
                 value = '********';
             }
             logger.info('Setting %s to environment value: %s', key, value);
--- a/lib/commonConfig.js
+++ b/lib/commonConfig.js
@@ -34,6 +34,9 @@ var config = {},
     commandRegistry,
     securityService;
 
+const secrets = require('@cloudreach/docker-secrets');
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     "watch": "watch 'npm test && npm run lint' ./lib ./test"
   },
   "dependencies": {
+    "@cloudreach/docker-secrets": "^1.0.3",
```
fgalan, line 43:
```suggestion
    "@cloudreach/docker-secrets": "~1.0.3",
```

jason-fox: Fixed 1dd0f77

fgalan:
Note that IOTAs are not always built using docker. For instance, we have RPM-based deployments in which case we inject directly the env vars to the node IOTA process.

Are the changes in this PR compatible with that way of working? I cannot follow the exact detail, but it seems is based in `@cloudreach/docker-secrets` package (and the "docker" literal on the package name makes me wonder if this could work only in docker deployments :)

jason-fox:
There is no issue with using ordinary ENV variables or amending the `config.js` file.  Of course if you do this you are still passing usernames and passwords in *plaintext* which is not exactly best practice security-wise :smile: 👮 

If you look at `getSecretData()` [here](https://github.com/telefonicaid/iotagent-node-lib/blob/1dd0f7714eadb9cac2f1a69371586d45828cb86d/lib/commonConfig.js#L53) you can see it is only Docker secret file processing for `ENV`s with the `_FILE` suffix.

`@cloudreach/docker-secrets` just takes the path of the env with the `_FILE` suffix and reads the data from the file and places it into the env without the `_FILE` suffix - it does the same job as the old `entrypoint.sh` bash script.

jason-fox:
Incidentally, processing this PR *before* the Docker Hardening is still backwards compatible, since the deleted `entrypoint.sh` will capture and process the env with the `_FILE` suffix and then delete it *before* starting the node application.

lib/commonConfig.js
```diff
@@ -34,6 +34,9 @@ var config = {},
     commandRegistry,
```
fgalan, line 34: An entry in CHANGES_NEXT_RELEASE about the changes in this PR should be added.

AlvaroVega (approved):
LGTM

lib/commonConfig.js
jason-fox: Fixed 49512da

jason-fox:
@AlvaroVega @fgalan 

Once this is merged someone should do a quick `git commit --allow-empty` on the IoT Agent for JSON `master` branch to add the Docker secret support back in.

fgalan (approved):
LGTM

fgalan:
> Once this is merged someone should do a quick `git commit --allow-empty` on the IoT Agent for JSON `master` branch to add the Docker secret support back in.

`master` is a protected branch in this repo. So I gess I can do `git commit --allow-empty` in my local copy, but I guess it will fail if I try `git push master` after that.

Maybe this time (exceptionally) we can manually trigger the build at dockerhub.

PR #886, opened 2020-08-05
feat: allow to provide a list of JSON-LD @context

bobeal:
Even if the NGSI-LD specification encourages the use of a wrapper @context (it is only required when provided via a Link header), it is still informative and cases where contexts are provided as a list of URLs will happen.

This PR aims at handling this.

### DIFF
--- a/lib/commonConfig.js
+++ b/lib/commonConfig.js
@@ -160,7 +160,7 @@ function processEnvironmentVariables() {
         config.contextBroker.ngsiVersion = process.env.IOTA_CB_NGSI_VERSION;
     }
     if (process.env.IOTA_JSON_LD_CONTEXT){
-        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT.split(',');
+        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT.split(',').map(ctx => ctx.trim());
     }
 
     config.contextBroker.fallbackTenant =  process.env.IOTA_FALLBACK_TENANT ||  
--- a/lib/commonConfig.js
+++ b/lib/commonConfig.js
@@ -160,7 +160,7 @@ function processEnvironmentVariables() {
         config.contextBroker.ngsiVersion = process.env.IOTA_CB_NGSI_VERSION;
     }
     if (process.env.IOTA_JSON_LD_CONTEXT){
-        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT;
+        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT.split(';');
     }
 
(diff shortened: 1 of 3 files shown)


lib/commonConfig.js
```diff
@@ -160,7 +160,7 @@ function processEnvironmentVariables() {
         config.contextBroker.ngsiVersion = process.env.IOTA_CB_NGSI_VERSION;
     }
     if (process.env.IOTA_JSON_LD_CONTEXT){
-        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT;
+        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT.split(';');
```
fgalan, line 163: Is this the usual pattern for env vars which in config.js may contain arrays? It is documented in the usage of env vars in general (or in IOTA_JSON_LD_CONTEXT env var in particular, if it isn't a general pattern)?

fgalan:
LGTM from @jason-fox would be derivable in this PR.

lib/commonConfig.js
bobeal: I've just added a note in `doc/installationguide.md` explaining this for the IOTA_JSON_LD_CONTEXT environment variable.

I've not found a general pattern for such a use case. Some use a semicolon, others use a comma. And a few use some exotic patterns.

jason-fox: Keyrock and PEP Proxy already use comma `,` for  transforming strings to arrays so this should align with them.

jason-fox:
**LGTM ... but** before merging this please swap to  using comma-separated rather than semi colon separated.

bobeal:
> **LGTM ... but** before merging this please swap to using comma-separated rather than semi colon separated.

Ok, did the switch to comma separated list.

lib/commonConfig.js
```diff
@@ -160,7 +160,7 @@ function processEnvironmentVariables() {
         config.contextBroker.ngsiVersion = process.env.IOTA_CB_NGSI_VERSION;
     }
     if (process.env.IOTA_JSON_LD_CONTEXT){
-        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT.split(';');
+        config.contextBroker.jsonLdContext = process.env.IOTA_JSON_LD_CONTEXT.split(',');
```
jason-fox, line 163: This should be a regex split including optional whitepace e.g. `/,\s*/`, as you just know someone will attempt to provision using:

- `IOTA_JSON_LD_CONTEXT=http:/context1.json-ld, http:/context2.json-ld,  http:/context3json-ld`

Instead of:

- `IOTA_JSON_LD_CONTEXT=http:/context1.json-ld,http:/context2.json-ld, http:/context3json-ld`

And then blame the code for not finding `%20http:/context2.json-ld`

jason-fox:
@fgalan Please merge

Issue #889, opened 2020-09-03
false DEVICE_NOT_FOUND log when simply create a new device


AlvaroVega:
Provisioning a new iota device is logging as error, and then raising and releasing an alarm.

```
iot-iota-ul                 | time=2020-09-03T10:24:20.565Z | lvl=ERROR | corr=7c1c6691-929d-4bf2-a215-9dbfad12cbc3 | trans=7c1c6691-929d-4bf2-a215-9dbfad12cbc3 | op=IoTAgentNGSI.Alarms | srv=n/a | subsrv=n/a | msg=Raising [MONGO-ALARM]: {"name":"DEVICE_NOT_FOUND","message":"No device was found with id:dispR","code":404} | comp=IoTAgent
iot-iota-ul                 | time=2020-09-03T10:24:20.569Z | lvl=ERROR | corr=7c1c6691-929d-4bf2-a215-9dbfad12cbc3 | trans=7c1c6691-929d-4bf2-a215-9dbfad12cbc3 | op=IoTAgentNGSI.Alarms | srv=n/a | subsrv=n/a | msg=Releasing [MONGO-ALARM] | comp=IoTAgent
```
But is normal not found a previous device created; it's just part of looking algorithm before register 
https://github.com/telefonicaid/iotagent-node-lib/blob/5214081f94288fc8e45092d68e991097f54e2610/lib/services/devices/deviceService.js#L1146

Produced by:
https://github.com/telefonicaid/iotagent-node-lib/blob/5214081f94288fc8e45092d68e991097f54e2610/lib/services/devices/deviceRegistryMongoDB.js#L199

Intercepted and logged by 
https://github.com/telefonicaid/iotagent-node-lib/blob/5214081f94288fc8e45092d68e991097f54e2610/lib/services/common/alarmManagement.js#L37

According with
https://github.com/telefonicaid/iotagent-node-lib/blob/5214081f94288fc8e45092d68e991097f54e2610/lib/services/devices/deviceRegistryMongoDB.js#L330

AlvaroVega:
And the similar logs when create a new group device:

[...]

PR #891, opened 2020-09-03
do not raise alarm if device is not found at findOrCreate 

AlvaroVega:
fix issue https://github.com/telefonicaid/iotagent-node-lib/issues/889


fgalan (approved):
LGTM

PR #898, opened 2020-09-08
Bug fix - command result.

jason-fox:
When updating an actuator and then querying the result, the response from the IoT Agent was invalid.

The forwarded query:

```console
curl -iX GET  'http://localhost:4041/ngsi-ld/v1/entities/urn:ngsi-ld:Device:water001?attrs=on,off'
```

was returning an array:

```json
[
  {
    "type": "Device",
    "id": "urn:ngsi-ld:Device:water001",
    "on": {
      "type": "command",
      "value": ""
    },
    "off": {
      "type": "command",
      "value": ""
    }
  }
]
```

It should have returned a single entity:

```json
{
  "type": "Device",
  "id": "urn:ngsi-ld:Device:water001",
  "on": {
    "type": "command",
    "value": ""
  },
  "off": {
    "type": "command",
    "value": ""
  }
}
```

Note that currently only single entities (e.g. `/urn:ngsi-ld:Device:water001`) can be *requested* and returned, however the code to deal with multiple entities (e.g. `/?type=Device`)still exists, because the full definition of context-forwarding to actutators is remains to be fully defined. It is on the proposals for inclusion in the next ESTI specification.

### DIFF
--- a/lib/services/northBound/contextServer-NGSI-LD.js
+++ b/lib/services/northBound/contextServer-NGSI-LD.js
@@ -385,10 +385,10 @@ function handleQueryNgsiLD(req, res, next) {
                     callback(error);
                 } else if (innerDevice.count) {
                     callback(null, results);
-                } else if (Array.isArray(results) && results.length > 0) {
+                } else if (Array.isArray(results) && results.length > 1) {
                     callback(null, results);
                 } else {
-                    callback(null, results);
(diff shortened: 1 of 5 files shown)


kzangeli:
LGTM

fgalan (approved):
LGTM

PR #902, opened 2020-09-09
Fix merge conflicts again.

jason-fox:
Merge branch 'master' into feature/842_ngsi_ld - fix conflicts again.

### DIFF
--- a/lib/services/devices/deviceRegistryMongoDB.js
+++ b/lib/services/devices/deviceRegistryMongoDB.js
@@ -30,7 +30,7 @@ const errors = require('../../errors');
 const constants = require('../../constants');
 const Device = require('../../model/Device');
 const async = require('async');
-const context = {
+let context = {
     op: 'IoTAgentNGSI.MongoDBDeviceRegister'
 };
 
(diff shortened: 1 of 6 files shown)


lib/services/devices/deviceRegistryMongoDB.js
```diff
@@ -30,7 +30,7 @@ const errors = require('../../errors');
 const constants = require('../../constants');
 const Device = require('../../model/Device');
 const async = require('async');
-const context = {
+let context = {
```
jason-fox, line 33: This looks very strange - it is necessary from #901  - it looks like that PR is amending this constant - is this correct?

fgalan: Not sure... maybe @AlvaroVega could provide feedback on this.

AlvaroVega: I guess there is no problem no move from `const` to `let`

fgalan: So I understand it's correct and NTC

fgalan (approved):
LGTM

PR #912, opened 2020-09-17
Correct ENV variable name.

jason-fox:
It is called `IOTA_EXPLICIT_ATTRS` in the docs and elsewhere in the code.


lib/commonConfig.js
```diff
@@ -130,7 +130,7 @@ function processEnvironmentVariables() {
             'IOTA_IOTAM_PROTOCOL',
```
fgalan, line 130: The fix included in this PR should be mentioned in CHANGES_NEXT_RELEASE. Suggestion:

```
- Fix: IOTA_EXPLICIT_ATTRS env var was not working
```

jason-fox: Fix 0ab2b88

fgalan (approved):
LGTM

PR #914, opened 2020-09-18
Update commonConfig.js

AlvaroVega:


jason-fox:
On reflection I think it would be better to remove the secrets package all together  and add in a `try ... catch` this means if it still fails on permissions can recover from it. #915 

The problem with this solution will be when someone tries to use Docker secrets on a RedHat systems - it will still fail then.

The joys of using `npm` package eh? 😄

fgalan:
overpassed at the end by PR telefonicaid/iotagent-node-lib#915

PR #915, opened 2020-09-18
Fix secrets for RedHat systems.

jason-fox:
Remove   @cloudreach/docker-secrets replace with custom code.

- Add try/catch
- Make Secrets locations configurable

### DIFF
--- a/lib/commonConfig.js
+++ b/lib/commonConfig.js
@@ -34,8 +34,25 @@ var config = {},
     commandRegistry,
     securityService;
 
-const secrets = require('@cloudreach/docker-secrets');
+const fs   = require('fs');
 const path = require('path');
+const SECRETS_DIR =  process.env.SECRETS_DIR || '/run/secrets';
+const secrets = {};
+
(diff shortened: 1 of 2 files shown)


jason-fox:
@AlvaroVega - this would be my solution.

jason-fox:
From: https://docs.docker.com/engine/swarm/secrets/

> When you grant a newly-created or running service access to a secret, the decrypted secret is mounted into the container in an in-memory filesystem. The location of the mount point within the container defaults to `/run/secrets/<secret_name>` in Linux containers, or `C:\ProgramData\Docker\secrets` in Windows containers. You can specify a custom location in Docker 17.06 and higher.

From: https://docs.docker.com/compose/compose-file/

> The short syntax variant only specifies the secret name. This grants the container access to the secret and mounts it at `/run/secrets/<secret_name>` within the container. The source name and destination mountpoint are both set to the secret name.

fgalan (approved):
LGTM

PR #919, opened 2020-09-28
Task/add from log

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/918

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,2 +1,3 @@
+Add: include from in log context (#918)
 Fix: Static attributes from service not applied if device has static attributes (#757)
 Move Docker secret support inside the Node Application - remove Entrypoint (#885)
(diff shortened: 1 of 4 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+Add: include from in log context (#918)
```
fgalan, line 1: Do we have some document in this repo describing the log system? In positive case, it should be updated due to ths PR. In negative case, this is NTC.

AlvaroVega: Not sure you mean something like this: https://github.com/telefonicaid/iotagent-node-lib/blob/96e12e12847dda7afd8cf6fb7af243a0ca9970c2/doc/operations.md#logs

fgalan: NTC

fgalan (approved):
LGTM

Issue #675, opened 2018-09-30
Attribute Default Value on Northbound side


gabrieledeluca:
During initialization of OCB entities, the library set some default attribute values based on type field. Here a snippet from _deviceService.js_ file:

`function getInitialValueForType(type) {
    switch (type) {
        case constants.LOCATION_TYPE:
            return constants.LOCATION_DEFAULT;
        case constants.DATETIME_TYPE:
            return constants.DATETIME_DEFAULT;
        default:
            return constants.ATTRIBUTE_DEFAULT;
    }
}`

The value of _constants.ATTRIBUTE_DEFAULT_ now is an empty string, what do you think about null value? In my opinion, semantically null is better: we are sensing something, so an empty string could be confused as real measure...

fgalan:
null would be probably semantically beterr... but, what about backward compatibilty? Existing clients of context information provided by IOTAgents may be using empty string to detect defaulta and chaging it to null could break them.

PR #920, opened 2020-09-29
Add lax validation mode for OPC-UA IoT Agent

jason-fox:
- Add IOTA_RELAX_TEMPLATE_VALIDATION env.
- Add switch for Lax/Strict ObjectId templating
- Amend documentation.

@gabrieledeluca - would this work for you?

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,4 +1,4 @@
 Fix: Static attributes from service not applied if device has static attributes (#757)
 Move Docker secret support inside the Node Application - remove Entrypoint (#885)
 Fix: IOTA_EXPLICIT_ATTRS env var was not working
-Add lax validation mode for OPC-UA IoT Agent using IOTA_RELAX_TEMPLATE_VALIDATION (#920)
+Add lax validation mode using IOTA_RELAX_TEMPLATE_VALIDATION (#920)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,3 +1,4 @@
 Fix: Static attributes from service not applied if device has static attributes (#757)
 Move Docker secret support inside the Node Application - remove Entrypoint (#885)
 Fix: IOTA_EXPLICIT_ATTRS env var was not working
+Add lax validation mode for OPC-UA IoT Agent using IOTA_RELAX_TEMPLATE_VALIDATION (#920)
--- a/doc/installationguide.md
+++ b/doc/installationguide.md
@@ -238,7 +238,11 @@ used for the same purpose. For instance:
 
 -   **explicitAttrs**: if this flag is activated, only provisioned attributes will be processed to Context Broker.
     This flag is overwritten by `explicitAttrs` flag in group or device provision.
--   **relaxTemplateValidation**: if this flag is activated, `objectIds` for incoming devices are not validated and may include characters such as semi-colons.
+-   **relaxTemplateValidation**: if this flag is activated, `objectId` attributes for incoming devices are not validated,
+    and may exceptionally include characters (such as semi-colons) which are
+    [forbidden](https://fiware-orion.readthedocs.io/en/master/user/forbidden_characters/index.html) according to the NGSI
+    specification.  When provisioning devices, it is necessary that the developer provides valid  `objectId`-`name` mappings
+    whenever relaxed mode is used, to prevent the consumption of forbidden characters.
 
 ### Configuration using environment variables
 
--- a/doc/installationguide.md
+++ b/doc/installationguide.md
@@ -233,11 +233,12 @@ used for the same purpose. For instance:
     [Cluster](https://nodejs.org/api/cluster.html) module in Node.js and
     [this section](howto.md#iot-agent-in-multi-thread-mode) of the library documentation.
 -   **defaultExpressionLanguage**: the default expression language used to
-    compute expressions, possible values are: `legacy` or `jexl`. When not set or 
+    compute expressions, possible values are: `legacy` or `jexl`. When not set or
     wrongly set, `legacy` is used as default value.
 
--   **explicitAttrs**: if this flag is activated, only provisioned attributes will be processed to Context Broker. 
+-   **explicitAttrs**: if this flag is activated, only provisioned attributes will be processed to Context Broker.
     This flag is overwritten by `explicitAttrs` flag in group or device provision.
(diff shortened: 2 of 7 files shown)


lib/services/northBound/deviceProvisioningServer.js
```diff
@@ -65,6 +66,15 @@ var async = require('async'),
         explicitAttrs: 'explicitAttrs'
     };
 
+
+if(config.relaxTemplateValidation){
```
fgalan, line 70: Code style:

```suggestion
if (config.relaxTemplateValidation) {
```

jason-fox: If only there was a way of dealing with these whitespace issues automatically? https://github.com/telefonicaid/iotagent-node-lib/pull/831

jason-fox: Fixed 62239f2

lib/services/northBound/deviceProvisioningServer.js
```diff
@@ -65,6 +66,15 @@ var async = require('async'),
         explicitAttrs: 'explicitAttrs'
     };
 
+
+if (config.relaxTemplateValidation) {
```
fgalan, line 70: Functionality added by this PR (along with a mention to the new env var) should be added to CHANGES_NEXT_RELEASE

jason-fox: Fixed cb1401e

fgalan:
Let's wait for @gabrieledeluca on this before merging.

jason-fox:
The OPC-UA fork diverges because of two issues. This fix would provide a viable workaround for #671 , but it does not address #675 - the latter was not addressed as it _may_ have backwards compatibility issues. 

If #675 is essential to operation I guess another flag will be necessary.

gabrieledeluca:
@jason-fox we created a pull request to your fork to fix a bug ([https://github.com/jason-fox/iotagent-node-lib/pull/1](https://github.com/jason-fox/iotagent-node-lib/pull/1)) :) 
thank you both

jason-fox:
> we created a pull request to your fork to fix a bug

Fixed 6a732ec1

lib/commonConfig.js
```diff
@@ -461,6 +462,11 @@ function processEnvironmentVariables() {
         config.defaultExpressionLanguage =
             process.env.IOTA_DEFAULT_EXPRESSION_LANGUAGE;
     }
+    if(process.env.IOTA_RELAX_TEMPLATE_VALIDATION){
+        config.relaxTemplateValidation = process.env.IOTA_RELAX_TEMPLATE_VALIDATION === 'true';
+    } else {
```
jason-fox, line 467: I don't think this second clause is necessary - if `config.relaxTemplateValidation` doesn't exist it will be falsy anyway.

manfredipist: Hi Jason, yes you're absolutely right, i just didn't want to leave an undefined value for future comparisons.

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
 Fix: Static attributes from service not applied if device has static attributes (#757)
 Move Docker secret support inside the Node Application - remove Entrypoint (#885)
 Fix: IOTA_EXPLICIT_ATTRS env var was not working
+Add lax validation mode for OPC-UA IoT Agent using IOTA_RELAX_TEMPLATE_VALIDATION (#920)
```
AlvaroVega, line 4: But the new parameter could be used by any iotagent, not just opc-ua iotagent, could be?

doc/installationguide.md
```diff
@@ -233,11 +233,12 @@ used for the same purpose. For instance:
...
     [this section](howto.md#iot-agent-in-multi-thread-mode) of the library documentation.
 -   **defaultExpressionLanguage**: the default expression language used to
-    compute expressions, possible values are: `legacy` or `jexl`. When not set or 
+    compute expressions, possible values are: `legacy` or `jexl`. When not set or
     wrongly set, `legacy` is used as default value.
 
--   **explicitAttrs**: if this flag is activated, only provisioned attributes will be processed to Context Broker. 
+-   **explicitAttrs**: if this flag is activated, only provisioned attributes will be processed to Context Broker.
     This flag is overwritten by `explicitAttrs` flag in group or device provision.
+-   **relaxTemplateValidation**: if this flag is activated, `objectIds` for incoming devices are not validated and may include characters such as semi-colons.
```
AlvaroVega, line 241: Orion ContextBroker is fine in that case?

fgalan: Good point... `;` is forbidden char in Orion https://fiware-orion.readthedocs.io/en/master/user/forbidden_characters/index.html.

jason-fox: But this is `objectId` which is relaxed - the ContextBroker uses the mapped `name` instead.

gabrieledeluca: Hi folks, I agree with @jason-fox , the _objectId_ is related to the devices (southbound side), it means it shouldn't be used for the NGSI mapping (northbound side). In a nutshell, we cannot give limitations or constraints to southbound IDs! Obviously, OCB forbidden chars should be taken in care in the _name_ field validation.

CHANGES_NEXT_RELEASE
fgalan: From a language point of view of course you are right ;) However, it's a matter of style. We try to describe the features in the changelog of the library in a IOTA-neutral way.

So, suggestion:

```suggestion
Add lax validation mode using IOTA_RELAX_TEMPLATE_VALIDATION (#920)
```

doc/installationguide.md
fgalan: Maybe that should be remarked in the documenation? Something like this:

```suggestion
-   **relaxTemplateValidation**: if this flag is activated, `objectIds` for incoming devices are not validated and may include characters such as semi-colons. Note that some characters are [forbidden in the CB](https://fiware-orion.readthedocs.io/en/master/user/forbidden_characters/index.html) so if you are using some of them in `objectId` in relax mode, take into account you need to use mappings using `name`s that doesn't have such forbidden chars.
```

Of course other wording expressing the same is also valid.

jason-fox: Fixed c92926f - I've reworded and clarified the text.

fgalan (approved):
LGTM. Thanks!

LGTM from @gabrieledeluca and confirmation that this PR overpasses old PR https://github.com/telefonicaid/iotagent-node-lib/pull/674 should be needed.

gabrieledeluca:
we can confirm it solved! thanks!

Issue #842, opened 2020-02-07
Adding NGSI-LD support


jason-fox:
It is necessary to add support for NGSI-LD to all IoT Agents. This is a large amount of work and needs to be split according. This issue is to discuss how the code can be incorporated and link the relevant PRs.

The main issue here is that there is a lot of work to review, so I've tried to split it into smaller PRs

1) [x] #839   - NGSI-LD relies very heavily on **metadata support** (a.k.a. Properties of Properties) - whilst adding the code it was noticed that aliasing didn't work nicely with metadata. Effectively this is a bug fix.
2)  [x] #840  - This is a simple Grep to be able to distinguish between NGSI-v2 and NGSI-LD tests. Very minor change across **lots** of files - I split it out to reduce the size of the PR
3)  [x] #841  - This is the main body of work for measures support
4)  [x] #843 - Temporal and GeoProperty support
5)  [x] #847 - Add NGSI-LD Multimeasures support
6)  [x] #848 - Add NGSI-LD Command support
7)  [x] #850 - Update the NGSI interactions to use ES-6 

### Dependencies

- #841 is dependent upon #839 ... and this is just the first **island of stability** - there is more to code.
- #843 is dependent upon #841  
- #847  is dependent upon #843 
- #848  is dependent upon #847 - this should now be basically usable (assuming the Context Broker interfaces are stable)
- #850 is just a simple automatic update of the files within `/services/devices/`, `/services/ngsi/` and `/services/northBound` to properly use `let` and `const` - May as well use this opportunity.

jason-fox:
Things which haven't been done yet (which I'd prefer to raise as separate smaller PRs)

- GeoProperty support
- Proper `unitCode`  metadata 
- Map `TimeInstant` onto `observedAt` metadata 
- Actuator Support (commands and results) - relies on Context Forwarding PATCH which is still in progress.
- Lazy Attributes  - relies on Context Forwarding GET which is still in progress.
- Documentation

The thing is that this is  still experimental, but would be **massive** if it is all collected into  one PR.

Basically need to discuss how this can be added into the codebase in a timely manner.

jason-fox:
The following repo provides a test harness for the NGSI-LD support.

https://github.com/jason-fox/tutorials.IoT-Agent/tree/test

To run the test harness just type:

``` console
./services start
```

[...]

PR #924, opened 2020-10-21
feat: use update mode of batch upsert when updating NGSI-LD entities

bobeal:
Hi all,

As discussed in https://github.com/telefonicaid/iotagent-node-lib/pull/888 (with a lot of delay, sorry for that), here is the PR to activate update mode when calling the batch upsert NGSI-LD API to update entities.

Default mode for batch upsert is to replace entities, which does not exactly match with what we want to do (basically updating a value measured by a sensor).

### DIFF
--- a/lib/services/devices/devices-NGSI-LD.js
+++ b/lib/services/devices/devices-NGSI-LD.js
@@ -213,7 +213,7 @@ function createInitialEntityNgsiLD(deviceData, newDevice, callback) {
  */
 function updateEntityNgsiLD(deviceData, updatedDevice, callback) {
     const options = {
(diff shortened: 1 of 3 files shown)


fgalan:
From my point of view, the PR is ok, but maybe feedback for other people more involved in NGSI-LD (maybe @jason-fox) would be desirable.

From a CHANGES_NEXT_RELEASE perspective maybe this PR should add a sub-bullet on the `Add basic NGSI-LD support as experimental feature (#842)` section.

jason-fox:
LGTM

fgalan (approved):
LGTM

PR #939, opened 2020-11-20
[WIP] use constant instead a value

AlvaroVega:
default value is 

https://github.com/telefonicaid/iotagent-node-lib/blob/510ae3713008761fa59f64b932a2641b449cd3ef/lib/constants.js#L51


lib/services/devices/deviceService.js
```diff
@@ -256,7 +256,7 @@ function formatCommandsNgsi2(originalVector) {
             };
```
fgalan, line 256: Should we add something to CNR? Something about defaults was previously wrongly set or something like that...

fgalan: (I didn't realize about the [WIP]... sorry)

Issue #938, opened 2020-11-19
IOTA creates attribute with blank value when there is a mapping


manucarrace:
When a device is provisioned with a mapping (e.g. modify the name of a device attribute) the IoTA is creating the entity attribute in Context Broker with value " ".

This value (" ") is provoking e2e errors in database since some attributes are defined as numeric in database.

Moreover, it is not clear if the attibute should be created in the device provision. It sounds more sensible that the attribute to be created with a received measure.

Besides, if a measure is received with several attributes, but the mapped attribute is not present in the measure, the attribute should not be updated in the Context Broker.

AlvaroVega:
Maybe we can achieve the desired behaviour using "explicitAttrs" device provision option:
`
optional boolean value, to support selective ignore of measures so that IOTA doesn’t progress. If not specified default is false.
`

AlvaroVega:
default value is handled by a constant: https://github.com/telefonicaid/iotagent-node-lib/pull/939

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/942. Probably this issue could be closed.

AlvaroVega:
Changes in iotas (tests) and e2e tests are also expected before merge branch task/default_attribute_value_null_instead_blank
So, in the mean time will be reverted: https://github.com/telefonicaid/iotagent-node-lib/pull/945

AlvaroVega:
fixed n PR https://github.com/telefonicaid/iotagent-node-lib/pull/946

fgalan:
> fixed n PR #946

That PR has been merged. Mabye this issue can be closed now? Or is there something else pending?

PR #942, opened 2020-11-24
Task/default attribute value null instead blank

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/938

No changes where done in default command attributes (info, status, ..)


fgalan (approved):
LGTM

Probably an old leftover from NGSIv1 days (NGSIv1 doesn't allow null) nicely solved with this PR :)

PR #948, opened 2020-11-26
REMOVE unneeded file

fgalan:
It seems this file was re-introduced by PR #849.

As far as I understand the file is no longer in use (as we changed from jshint to eslint recently) and should be removed, isn't it @jason-fox ?


jason-fox:
Yes, remove it.

AlvaroVega (approved):
LGTM

PR #975, opened 2021-01-20
add device.apikey to provisioningAPITranslation

AlvaroVega:
Currently a GET device does not retrieve apikey if was provisioned at device level
This PR returns apikey with device provisioning data if was provisioned.

issue https://github.com/telefonicaid/iotagent-node-lib/issues/977


fgalan (approved):
LGTM

PR #980, opened 2021-01-25
Feature/null check

jason-fox:
Potential fix for https://github.com/telefonicaid/iotagent-json/issues/523

### DIFF
--- a/lib/services/ngsi/entities-NGSI-LD.js
+++ b/lib/services/ngsi/entities-NGSI-LD.js
@@ -71,7 +71,14 @@ function convertNGSIv2ToLD(attr) {
     switch (attr.type.toLowerCase()) {
         // Properties
         case 'property':
+            if (obj.value === null){
+                obj.value = NGSI_LD_NULL;
+            }
+            break;
         case 'string':
+            if (obj.value === null){
+                obj.value = '';
+            }
             break;
 
         // Other Native JSON Types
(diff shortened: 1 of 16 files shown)


lib/services/ngsi/entities-NGSI-LD.js
```diff
@@ -68,93 +68,98 @@ function valueOfOrNull(value) {
```
fgalan, line 68: I understand this PR doesn't need to add any entry to CNR file given the existing "Add: basic NGSI-LD support as experimental feature (#842)" already covers it, right?

jason-fox: Correct.

fgalan: NTC

jason-fox:
I've completed updating the test cases. The remaining error is just a **Coveralls** rounding error.

fgalan:
> I've completed updating the test cases. The remaining error is just a Coveralls rounding error.

Maybe coveralls can be configured with some "tolerance" degree? I mean, for instance, if the difference is less than 0.05% don't mark it as fail.

fgalan (approved):
LGTM

jason-fox:
> > I've completed updating the test cases. The remaining error is just a Coveralls rounding error.
> 
> Maybe coveralls can be configured with some "tolerance" degree? I mean, for instance, if the difference is less than 0.05% don't mark it as fail.

You can do this - just read https://docs.coveralls.io/coveralls-notifications

Go to https://coveralls.io/github//telefonicaid/iotagent-node-lib/settings

Screenshot 2021-01-26 at 17 37 25

PR #984, opened 2021-02-02
Fix #983 - add options=update

### LINKED ISSUE #983: Attributes/commands removed after device provisioning
If I run iotagent-ul or iotagent-json over NGSI-LD and provisioning a device with attributes and/or commands. When the device sends the first message, NGSI-LD context broker, for instance orion-LD removes all attributes/commands from the broker except are included in this first message... seems iotagent-node-lib/lib/services/ngsi/entities-NGSI-LD.js files is wrong and "/ngsi-ld/v1/entityOperations/upsert" must be modified to "/ngsi-ld/v1/entityOperations/upsert/?options=update".

jason-fox:
Use   `/ngsi-ld/v1/entityOperations/upsert/?options=update` consistently

One line code change plus 15 updated test files


fgalan (approved):
LGTM

@jason-fox somebody else should review and provide LGTM? Or should I merge now?

jason-fox:
You may as well merge.

PR #985, opened 2021-02-05
Check command count

jason-fox:
Add test for multiple command handler calls for NGSI-v2 and NGSI-LD


jason-fox:
The issue was not caught due to the unit test - I've swapped the `boolean` for an incrementing count and indeed you can see that the command handler was called twice.

There probably is a better way than just placing the body in an array for the mapping function, but it works and seems a logical fix due to the error. That is a req.body with two attributes was firing twice rather than an array with one element firing once.

fgalan (approved):
LGTM

Issue #770, opened 2019-04-05
Error when updating associated entity in a device


manucarrace:
When updating a device, if the associated entity is changed, then the measures received by the IOTA can not reach the Context Broker, since the entity does not exist.

manucarrace:
After some investigations we have seen the following.

When the IoTAgent receives a measure it makes a request to Context Broker like this:
http://10.0.0.2:1026/v2/entities/AV_MEDVO_CA_0001/attrs

This requests only updates the value of the attribute, so, if the entity doesn't exist the Context Broker will respond with an error.

This issue would be solved by creating the new entity when updating the device.

fgalan:
Just to add some complementary information to the @manucarrace report:

* We have done tests with IOTA-JSON with MQTT transport but it seems a common behaviour in the library, so probably no matter the agent no matter the transport it will be the same.
* The operation that IOTA does on CB is `POST /v2/entities/<devEnt>/attrs`. Thus, the error happens if `<devEnt>` doesn't exist. Curiously, if expression plugin is in place, the operation used in `POST /v2/op/update` with actionType: "append" so the problem doesn't happen (a new entity is created).
* I fully agree in the suggested solution. It would be a metter of modifying the `PUT /iot/device/<dev>` operation logic to include new entity creation (actually, an upsert to avoid colission if the entity already exists) in the workflow.

manucarrace:
Searching in the documentation I've found the following:

"PUT /iot/devices/:deviceId
Changes the stored values for the device with the provided Device payload. Neither the name, the type nor the ID of the device can be changed using this method (as they are used to link the already created entities in the CB to the information in the device). Service and servicepath, being taken from the headers, can't be changed also."

If this is the desired behaviour, a change in the entity_name should be forbidden by APIn and it should return an error. 
If the desired behaviour is that a device could change its asociated entity (is sounds sensitive to me) then a change in line with Fermin solution would be nice.

fgalan:
> If this is the desired behaviour, a change in the entity_name should be forbidden by APIn and it should return an error.

Probably this is the easiest solution.

Madhu-NEC:
Hi @fgalan , @manucarrace ,
I have started working on this issue. Could you please assign this issue to me?

**[Current Behaviour]**
**device_id**: It does not get updated but does not return an error.
**entity_name**: It gets updated.
**entity_type**: It gets updated.
 
**[Expected Behaviour]**
Below is my understanding for this issue.
[...]

PR #988, opened 2021-02-12
update entity_type and entity_name of devices

Madhu-NEC:
Fixes issue #770

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,3 +1,4 @@
+Fix: update type and name of device (#770)
 Fix: ensure registrationId is updated when defice with commands is updated 
 Add: return apikey in GET device if device was provisioned with explicit apikey (#977)
 Add: allow update device `apikey` using Update (put) API (#978)
--- a/lib/services/devices/devices-NGSI-LD.js
+++ b/lib/services/devices/devices-NGSI-LD.js
@@ -275,7 +275,7 @@ function updateEntityNgsiLD(deviceData, updatedDevice, callback) {
  *
  * @param {Object} deviceObj                    Object with all the device information (mandatory).
  */
-function updateRegisterDeviceNgsiLD(deviceObj, callback) {
+function updateRegisterDeviceNgsiLD(deviceObj, typeNameUpdated, callback) {
     if (!deviceObj.id || !deviceObj.type) {
         callback(new errors.MissingAttributes('Id or device missing'));
         return;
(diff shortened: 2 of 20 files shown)


lib/services/northBound/deviceProvisioningServer.js
```diff
@@ -316,7 +316,7 @@ function handleRemoveDevice(req, res, next) {
  * This middleware handles updates in the provisioning devices. The only attribute
  */
 function handleUpdateDevice(req, res, next) {
-    if (req.body.deviceId) {
+    if (req.body.device_id) {
```
fgalan, line 319: Coding style is to use camelCase not snake_case. Except if there is some important reason (which one?), please use deviceId

test/unit/ngsiv2/examples/contextAvailabilityRequests/updateIoTAgent1.json
```diff
@@ -3,7 +3,7 @@
         "entities": [
             {
                 "type": "Light",
-                "id": "light1"
+                "id": "Light:light1"
```
fgalan, line 6: Does the change in this existing test mean a change of existing behavior?

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+Fix: update type and name of device (#770)
```
fgalan, line 1: Maybe a better description whould be this one?

```suggestion
Fix: PUT /iot/devices/:deviceId operation to implement proper update of entity_name and entity_type (#770)
```

Please, check if I'm right before accepting it (I may be wrong)

lib/services/devices/devices-NGSI-LD.js
```diff
@@ -275,7 +275,7 @@ function updateEntityNgsiLD(deviceData, updatedDevice, callback) {
  *
  * @param {Object} deviceObj                    Object with all the device information (mandatory).
  */
-function updateRegisterDeviceNgsiLD(deviceObj, callback) {
+function updateRegisterDeviceNgsiLD(deviceObj, typeNameUpdated, callback) {
```
fgalan, line 278: I understand `typeNameUpdate` means "type and name [=id] of the entity update". Maybe a better name could be entityUpdated or entityInfoUpdated ("typeName" is less clear, from my point of view).

lib/services/northBound/deviceProvisioningServer.js
Madhu-NEC: As per the https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/templates/updateDevice.json#L7 , the property used for device id is device_id. So, I have changed it to device_id.

test/unit/ngsiv2/examples/contextAvailabilityRequests/updateIoTAgent1.json
Madhu-NEC: As per the existing behaviour entity_name can't be changed, so this id was not into effect. Now, if the entity_name is allowed to get updated then it should be same as per the standards. It will not change the existing behaviour.

CHANGES_NEXT_RELEASE
Madhu-NEC: Fixed in b7817e0

lib/services/devices/devices-NGSI-LD.js
Madhu-NEC: Fixed in 2cd4d32

Madhu-NEC:
Hi @fgalan, 
Thanks for your support.
I have updated the PR as per the suggestions shared.

lib/services/northBound/deviceProvisioningServer.js
AlvaroVega: Maybe `device_id` is imposed by API translation  and should be still kept
https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/northBound/deviceProvisioningServer.js#L197

Madhu-NEC: > Maybe `deviceId` is imposed by API translation and should be still kept
> https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/northBound/deviceProvisioningServer.js#L197

Hi @AlvaroVega ,
As per my understanding, here the device_id is read from request's body without imposing **toProvisioningAPIFormat(device)** then device_id should be same as per the template https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/templates/updateDevice.json#L7 . Please correct me if my understanding is wrong.

fgalan: I seems that the translate funciton is used only in `res.status(200).json(toProvisioningAPIFormat(device));`, i.e. at response time.

Thus probably you are right @Madhu-NEC and changing from `req.body.deviceId` to `req.body.device_id` here is a good fix.

test/unit/ngsiv2/examples/contextAvailabilityRequests/updateIoTAgent1.json
fgalan: NTC

fgalan (approved):
LGTM. Thanks for your contribution!

fgalan:
Please, solve the conflict in CHANGES_NEXT_RELEASE so this PR could be merged.

lib/services/northBound/deviceProvisioningServer.js
fgalan: NTC

Issue #989, opened 2021-02-16
enable multientity (+expression) in attribute group


AlvaroVega:

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/990

PR #990, opened 2021-02-17
check and set proper expression plugin to use in multientity

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/989


fgalan (approved):
LGTM

Issue #974, opened 2021-01-15
Proper use of supportedForwardingMode in registrations


fgalan:
With the implementation of `supportedForwardingMode` in Orion Context Broker (issue https://github.com/telefonicaid/fiware-orion/issues/3106), the NGSIv2 registrations done by the IOTA Library should be adapted the following way:

* Registrations corresponding to **lazy attributes** should use ` "supportedForwardingMode": "query"`
* Registrations corresponding to **commands** should use ` "supportedForwardingMode": "update"`

NGSIv1 and NGSI-LD will be out of the scope of this issue.

AlvaroVega:
Ngsi registers are handled in https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/devices/registrationUtils.js

fgalan:
> Ngsi registers are handled in https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/devices/registrationUtils.js

It seems IOTA does the registration in a single `POST /v2/registrations`. To take advantage of this feature, now we need to split into two ones, each one with a different supportedForwardingMode (query and update)

fgalan:
Some more findings:

1. Looking into the `sendRegistrationsNgsi2()` function I have found the fragment of the c

```
    const options = {
        url: cbHost + '/v2/registrations',
        method: 'POST',
        json: {
            dataProvided: {
                entities: [
                    {
                        type: deviceData.type,
                        id: String(deviceData.name)
                    }
                ],
                attrs: []
            },
            provider: {
                http: {
                    url: config.getConfig().providerUrl
                }
            }
        },
        headers: {
            'fiware-service': deviceData.service,
            'fiware-servicepath': deviceData.subservice
        }
    };
```

So a "wide" registration is being used (note the `attrs: []`). Not sure if at this point of the code we have access to lazy and commands attributes. Is that information being passed down to the function and accessible to build the registration payload at that point?

2. The registration ID is stored in devices collection in database as an string, i.e.:

```
"registrationId" : "5f984771106a3ac82b336f3e",
```

If we are going to use two registrations, the two strings need to be stored (either in an array, a sub-object or as an additional field, e.g. `registrationIdLazy")

AlvaroVega:
@fgalan is possible that lazy and commands are present at the same time for the same registration ?

fgalan:
> @fgalan is possible that lazy and commands are present at the same time for the same registration ?

[...]

PR #996, opened 2021-02-25
[WIP] add supportedForwardingMode

AlvaroVega:
related https://github.com/telefonicaid/iotagent-node-lib/issues/974
This pr is assumming that if lazy attrs there is not commands and otherwise (if commands  there is not lazy atts) so this is a partially fix


AlvaroVega:
There are many changes from this  PR was done, so this PR should be re-implemented again from the scratch

Issue #1007, opened 2021-03-23
provisioning a new device is still showing raise/releasing MONGO alarms


AlvaroVega:
time=2021-03-23T10:29:55.926Z | lvl=DEBUG | corr=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | trans=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | op=IoTAgentNGSI.GenericMiddlewares | from=n/a | srv=smartcity | subsrv=/ | msg=Request for path [/iot/devices] from [localhost:18082] | comp=IoTAgent
time=2021-03-23T10:29:55.926Z | lvl=DEBUG | corr=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | trans=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | op=IoTAgentNGSI.GenericMiddlewares | from=n/a | srv=smartcity | subsrv=/ | msg=Body:

{
    "devices": [
        {
            "device_id": "AIQ_A464",
            "entity_name": "IAQ_MA001",
            "entity_type": "IAQ",
            "static_attributes": [
                {
                    "value": "36.71749, -4.4743",
                    "type": "geo:point",
                    "name": "location"
                },
                {
                    "name": "address",
                    "type": "String",
                    "value": "Facultad"
                }
            ],
            "protocol": "IoTA-UL"
        }
    ]
}

 | comp=IoTAgent
time=2021-03-23T10:29:55.927Z | lvl=DEBUG | corr=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | trans=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | op=IoTAgentNGSI.DeviceProvisioning | from=n/a | srv=smartcity | subsrv=/ | msg=Handling device provisioning request. | comp=IoTAgent
time=2021-03-23T10:29:55.928Z | lvl=DEBUG | corr=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | trans=a0e5b0c0-c43b-4339-a6a5-063d522b0912 | op=IoTAgentNGSI.MongoDBGroupRegister | from=n/a | srv=n/a | subsrv=n/a | msg=Looking for group params ["service","subservice","type","apikey"] with queryObj {"service":"smartcity","subservice":"/","type":"IAQ"} | comp=IoTAgent
[...]

PR #1008, opened 2021-03-23
get silently checking duplicates

AlvaroVega:
related https://github.com/telefonicaid/iotagent-node-lib/issues/1007

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,4 +1,5 @@
-- Add: expose getConfigurationSilently to enable retrieve a configuration without raise a mongo alarm.
+- Add: use getDeviceSilently in checkDuplicates to avoid raise a false mongo alarm.
+- Add: expose getConfigurationSilently to enable retrieve a configuration without raise a false mongo alarm.
 - Add: db uri and options in mongo connection log INFO trace
 - Set Nodejs 12 as minimum version in packages.json (effectively removing Nodev10 from supported versions)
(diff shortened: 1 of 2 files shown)


AlvaroVega:
there is already a https://github.com/telefonicaid/iotagent-node-lib/blob/667b78ab331aeb8a0564593bcfc214770c361eb7/lib/services/devices/deviceService.js#L228

fgalan:
> Regarding with lint failure there is already a
> 
> https://github.com/telefonicaid/iotagent-node-lib/blob/667b78ab331aeb8a0564593bcfc214770c361eb7/lib/services/devices/deviceService.js#L228

@jason-fox you have more experience in the linter than us. Do you know why the linter raises an error when the disable-next comment is just above that line? For us is quite a mystery... :)

jason-fox:
```
 config.getRegistry().get(deviceObj.id, deviceObj.service, deviceObj.subservice, function (error, device) {
 ```
 
 You are not using `device` in the function below. However it can be useful to leave the full signature since it **may** prove useful should you need to refactor at some point. Better to stop eslint from complaining here.

**Comment by jason-fox** [comment]
If eslint is complaining, then maybe just move the comment closer to the "error"

```javascript
config.getRegistry().get(deviceObj.id, deviceObj.service, 
/* eslint-disable-no-unused-vars */
deviceObj.subservice, function (error, device) {
```

fgalan:
It works. Thanks for the feedback @jason-fox !

fgalan (approved):
LGTM

PR #1010, opened 2021-03-29
fix group command in device when entity_type is not equal group entity_type

AlvaroVega:
Fix for https://github.com/telefonicaid/iotagent-node-lib/issues/1011

Search by `apikey` should be enough since `apikey` is unique key in Group model. Ensure that also type match is not necessary.


fgalan (approved):
LGTM

Issue #1012, opened 2021-04-06
device attribute of type "Point" is translated to entity as geo:json 


AlvaroVega:
This is breaking previous client integrations 

Comes from https://github.com/telefonicaid/iotagent-node-lib/pull/854

Translation is done by  https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/ngsi/entities-NGSI-v2.js#L59

Possible Fix:  https://github.com/telefonicaid/iotagent-node-lib/pull/1013
 - [x] Need checks other types (like geo:point, etc)  to keep previous behaviour
 - [x] Need update tests
 - [x] Need update doc

geo types involved:
  geoproperty, point, geo:point, 
  linestring, geo:linestring, 
  polygon, geo:polygon,  
  multipoint, geo:multipoint, 
  multilinestring, geo:multilinestring, 
  multipolygon, geo:multipolygon

fgalan:
PR #1013 (for brach release/2.15.0) and PR #1014 (for branch master) provide a hotfix for this, but a definitive fix is needed, so this issue is kept open.

The definitive fix is only needed in master branch.

AlvaroVega:
related  https://github.com/telefonicaid/iotagent-node-lib/pull/1020

fgalan:
> related #1020

Reopening issue, while PR https://github.com/telefonicaid/iotagent-node-lib/pull/1020 is opened

PR #1020, opened 2021-04-12
Re-introduce GeoJSON conversion under a flag

jason-fox:
Reverts #1014 and #1018. Adds `autocastGeoJSON` to the config for use with NGSI-LD legacy systems.

- Updated code
- Added unit test
- Updated Docs


mapedraza:
Pending to define Group provision and device provision JSON payload how to enable those casting functionalities instead of having a global env flag

jason-fox:
> Pending to define Group provision and device provision JSON payload how to enable those casting functionalities instead of having a global env flag

@mapedraza - the global flag has been removed and `autocastGeoJSON = true` added to the attribute metadata. I think this is now ready for review.

doc/advanced-topics.md
```diff
@@ -102,6 +103,27 @@ accepted. Provided the `type` is provisioned correctly, the `value` may be defin
...
+    "entity_type": "GPS",
+    "resource":    "/iot/d",
+    "protocol":    "PDI-IoTA-JSON",
+..etc
+    "attributes": [
+        {
+            "name": "observationSpace",
+            "type": "Polygon",
+            "autocastGeoJSON" : "true"
+        }
```
fgalan, line 122: I think the following syntax would be better:

```
        {
            "name": "observationSpace",
            "type": "Polygon",
            "autocast" : true
        }
```

So the way in which the autocast is done depends on the `type`. This way we can extend the functionality in the future to autocast other types. E.g. to autocast to boolean (e.g. `"true"` -> `true`, `0` -> `false`):


```
        {
            "name": "observationSpaceEnabled",
            "type": "Boolean",
            "autocast" : true
        }
```

Maybe it would even better to have a `autocastType` to cover cases in which the NGSI type need to be customized by the user, eg:

```
        {
            "name": "observationSpace",
            "type": "MyCustomType",
            "autocastType": "Polygon",
            "autocast" : true
        }
```

The `autocastType` could be optional and, in the case it is ommited, `type` is used.

doc/advanced-topics.md
```diff
@@ -102,6 +103,27 @@ accepted. Provided the `type` is provisioned correctly, the `value` may be defin
 }
 ```
 
+Because GeoJSON types (e.g. `Point`, `LineString` etc.) are native types in **NGSI-LD**, automatic GeoJSON conversion is switched on for NGSI-LD by default.
+
+With **NGSI-v2**, for backwards compatibility reasons, automatic GeoJSON conversion for types other than `geo:json` is turned off by default.
+Add the `autocastGeoJSON` configuration to the attribute to enable GeoJSON conversion. Each GeoJSON attribute can be provisioned as shown:
+
+```json
```
fgalan, line 111: This example in documentation is for a configuration group. However, the templates modified in this PR are only about devices:

The behavior should be as follow:

imagen

* It can be enabled/disabled at configuration group level
* It can be enabled/disabled at device level. If not defined at device level, the config group setting is used (and if not defined at config group level, then assume false by default)

doc/advanced-topics.md
```diff
@@ -70,10 +70,10 @@ curl http://${KEYSTONE_HOST}/v3/OS-TRUST/trusts \
 Apart from the generation of the trust, the use of secured Context Brokers should be transparent to the user of the IoT
```
fgalan, line 70: I understand this PR provides "the definitive fix" mentioned in https://github.com/telefonicaid/iotagent-node-lib/issues/1012#issuecomment-814014235

Thus, an entry in CHANGES_NEXT_RELEASE relating issue #1012 should be included.

PR #1034, opened 2021-05-10
remove cast by type after apply JEXL expression

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/1036

- [x] Update related tests

The following tests are now not possible:
- Update for an integer attribute with string expression
- Update for a Float attribute with string expression
- Update for a Null attribute with arithmetic expression
- Update for a Null attribute with string expression
- Update for a Boolean attribute with arithmetic expression
- Update for a Boolean attribute with string expression


lib/plugins/jexlParser.js
```diff
@@ -125,26 +125,6 @@ function expressionApplier(context, typeInformation) {
         }
 
         newAttribute.value = applyExpression(attribute.expression, context, typeInformation);
-
```
fgalan, line 128: Is there any piece of documentation about this autocast feature that should be also removed/modified?

test/unit/expressions/expressionCombinedTransformations-test.js
```diff
@@ -192,7 +192,7 @@ describe('Combine Jexl and legacy expressions (default JEXL)', function () {
                 .matchHeader('fiware-servicepath', 'gardens')
```
fgalan, line 192: We have also a variant of this tests for ngsiv2

test/unit/ngsiv2/expressions/jexlBasedTransformations-test.js

That file doesn't need modifications? (looking to CI results, it seems it doesn't, but I wonder why...)

test/unit/ngsiv2/expressions/jexlBasedTransformations-test.js
```diff
@@ -385,40 +385,41 @@ describe('Java expression language (JEXL) based transformations plugin', functio
         });
     });
 
-    describe('When an update comes for attributes with string expression and type integer', function () {
```
fgalan, line 388: We have also a variant of this tests for ngsiv2

test/unit/ngsiv2/expressions/jexlBasedTransformations-test.js

That file doesn't need modifications? (looking to CI results, it seems it doesn't, but I wonder why...)

test/unit/ngsiv2/expressions/jexlBasedTransformations-test.js
```diff
@@ -385,40 +385,41 @@ describe('Java expression language (JEXL) based transformations plugin', functio
...
-
-        it('should apply the expression before sending the values', function (done) {
-            iotAgentLib.update('ws1', 'WeatherStationMultiple', '', values, function (error) {
-                should.not.exist(error);
-                contextBrokerMock.done();
-                done();
-            });
-        });
-    });
+    // Deleted test by Issue #1034
```
fgalan, line 388: If the "autocast after jexl processing" removal is definitive, maybe it's better to remove these tests instead of commenting. What do you think?

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Fix: Do not transform attribute value using attribute type after apply expression plugin JEXL (#1034)
```
fgalan, line 1: #1034 is the number of the PR itself. I understand that in this case we don't have any issue, so the PR is cited, right?

lib/plugins/jexlParser.js
AlvaroVega: There is nothing about this autocast in https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/expressionLanguage.md#jexl-based-transformations

test/unit/expressions/expressionCombinedTransformations-test.js
AlvaroVega: expressionCombinedTransformations-test.js includes tests which legacy and jexl expression, but not any of now not allowed cases:

    Update for an integer attribute with string expression
    Update for a Float attribute with string expression
    Update for a Null attribute with arithmetic expression
    Update for a Null attribute with string expression
    Update for a Boolean attribute with arithmetic expression
    Update for a Boolean attribute with string expression

test/unit/expressions/expressionCombinedTransformations-test.js
```diff
@@ -192,7 +192,7 @@ describe('Combine Jexl and legacy expressions (default JEXL)', function () {
...
                 .post(
                     '/v1/updateContext',
-                    utils.readExampleFile('./test/unit/examples/contextRequests/updateContextExpressionPlugin3.json')
+                    utils.readExampleFile('./test/unit/examples/contextRequests/updateContextExpressionPlugin3b.json')
```
fgalan, line 195: Instead of creating a new file (with the b suffix) I think it would be getter to edit the current ones. This way, the removal done in PR https://github.com/telefonicaid/iotagent-node-lib/pull/995 would be easy (as there files are currently taken into account for the removal).

CHANGES_NEXT_RELEASE
AlvaroVega: fixed with issue  number 1034

test/unit/expressions/expressionCombinedTransformations-test.js
AlvaroVega: IMHO is not possible edit current file, and should be splited to use in different tests. 
Renamed files done into 8f3ea370c13ebee1ec8c5b777523168af70f65f0

test/unit/ngsiv2/expressions/jexlBasedTransformations-test.js
fgalan: Fixed in 086a53f

CHANGES_NEXT_RELEASE
fgalan: Fixed in c6a1f70

test/unit/expressions/expressionCombinedTransformations-test.js
fgalan: Ok. I see (for instance https://github.com/telefonicaid/iotagent-node-lib/search?q=updateContextExpressionPlugin3).

I'll comment about the new files added in this PR on #995

fgalan: > I'll comment about the new files added in this PR on #995

Here: https://github.com/telefonicaid/iotagent-node-lib/pull/995#issuecomment-840368989

NTC

lib/plugins/jexlParser.js
fgalan: NTC

test/unit/expressions/expressionCombinedTransformations-test.js
fgalan: I have had a look into this with more detail.

In test/unit/expressions/expressionCombinedTransformations-test.js (NGSIv1 variant) the only difference between old files (e.g. test/unit/examples/contextRequests/updateContextExpressionPlugin3.json) and new files (test/unit/examples/contextRequests/updateContextExpressionPlugin8.json) is that the old uses `"value": "1040"` ("stringfied") and the new `"value": 1040` ("native").

On the other hand test/unit/ngsiv2/expressions/expressionCombinedTransformations-test.js (NGSIv2 variant) already uses the native way, e.g. see test/unit/ngsiv2/examples/contextRequests/updateContextExpressionPlugin1.json.

Thus, I think that explains that only test/unit/expressions/expressionCombinedTransformations-test.js is changed in this PR but not test/unit/ngsiv2/expressions/expressionCombinedTransformations-test.js

NTC

fgalan (approved):
LGTM

PR #1038, opened 2021-05-12
check entityId is valid after apply expression in multientity plugin

AlvaroVega:
jexl applyExpression could return NaN

fix for https://github.com/telefonicaid/iotagent-node-lib/issues/1039


fgalan (approved):
LGTM

PR #1042, opened 2021-05-18
Test only removal of NGSI-v1

jason-fox:
Attempt to reduce the scope of #995 - just updating tests


jason-fox:
Closing as discussed.

Issue #1045, opened 2021-05-24
explicitAttrs doesn't work for group provision


manucarrace:
In a device group provision, with explicitAttrs

```json
{
    "services": [
        {
            "apikey": "{{JSON_APIKEY}}",
            "entity_type": "device",
            "explicitAttrs": true,
            "attributes": [
                {
                    "object_id": "t",
                    "name": "temperature",
                    "type": "Number"
                },
                {
                    "object_id": "h",
                    "name": "humidity",
                    "type": "Number"
                }
            ],
            "protocol": [
                "IoTA-JSON"
            ]
        }
    ]
}
```

When you send an attribute not defined (e.g. "e"), it is delivered to Context Broker.
http://{{ENDPOINT_IOTA_JSON_HTTP}}/iot/json?i=dev2json&k={{JSON_APIKEY}}
```json
{
  "t": 35,
  "e": 34
}
```

Both attributes are present in the entity

However, this mechanism works properly in the device provision.

fgalan:
Maybe related issue: https://github.com/telefonicaid/iotagent-node-lib/issues/1016

fgalan:
After merging PR #1046 this issue could have been solved.

PR #1029, opened 2021-05-04
use expresionLanguage from group if not defined

AlvaroVega:
fix for https://github.com/telefonicaid/iotagent-node-lib/issues/1027

Issue #1016, opened 2021-04-08
explicitAttrs function not working properly?


SBlechmann:
Hey there,

as statet in #1003 with the new lib version I was able to get the autoprovision function working. 

I also tried to use the function of explicitAttrs. I created two devices, one with explicitAttrs=true and one with false. When sending data via MQTT to the respective devices using non-provisioned attributes both times attributes are created in the context broker.

Am I using the function wrong or is there a bug?

Thanks for your time!

`{
      "device_id": "tsensor001",
       ...[other stuff]...
      "transport": "MQTT",
      "attributes": [
        {
          "object_id": "t",
          "name": "temperature",
          "type": "Number"
        }
      ],
      "lazy": [],
      "commands": [],
      "static_attributes": [],
      "internal_attributes": [],
      "protocol": "PDI-IoTA-JSON",
      "explicitAttrs": true
    },`

fgalan:
It would be great if you could provide the exact sequence of steps to reproduce the problem (curl and mqtt_publish statements are preferred) since you start the agent with an empty database.

SBlechmann:
I'd reference back to #1003.

fgalan:
> I'd reference back to #1003.

Not sure about what you mean... Do you mean the sequence of step is *exactly the same* you describe at https://github.com/telefonicaid/iotagent-node-lib/issues/1003#issuecomment-817155239 ?

SBlechmann:
Yes, I have only conducted the tests described in #1003. 
The results shown there reflect the behaviour of explicitAttrs as well.

mapedraza:
After replicating what @SBlechmann was doing using iotagent-json:1.17.1, I got the following conclusions.

### Using NGSI V1
- with `explicitAttrs`=true on Config group JSON
   - `explicitAttrs`=true on device provision: New attribute is not included  - OK
   - `explicitAttrs`=false on device provision: New attribute is not included - NOK (Device parameter should overwrite group option)
   - `explicitAttrs` not present on device provision: New attribute is not included - OK
- with `explicitAttrs`=false on Config group JSON
   - New attributes are not included 


### Using NGSI V2 (`IOTA_CB_NGSI_VERSION=v2`)
- with `explicitAttrs`=true on Config group JSON
   - `explicitAttrs`=true on device provision: New attribute is included  - NOK
   - `explicitAttrs`=false on device provision: New attribute is included  - OK
[...]

Issue #1003, opened 2021-03-11
Autoprovision functionality not working as expected when configured at device/group level


kandarp-n:
Steps to reproduce:
1) Register a device with 'autoprovision' flag set to true:
```
{
"devices": [
   {
     "device_id":   "sensor001",
     "entity_name": "urn:ngsi-ld:Sensor:sensor001",
     "entity_type": "Sensor",
     "timezone":    "Europe/Berlin",
     "attributes": [
       { "object_id": "temperature", "name":"temperature", "type":"number"}
      ],
      **"autoprovision":true**
   }
 ]
}
```

2) Send measurement for this device with an extra measurement attribute. Sample measurement payload:
```
{
    "temperature":23.4,
    "humidity":40
}
```

**Expected result**: As "autoprovision" is "true" for this device,  the new attribute "humidity" should get created in the Orion entity with its measuremnt value. 
**Actual result**:  "humidity" attibute is not get created in Orion entity. HTTP status code 404 is returned. (Note: Measurement of "temperature" is getting updated in Orion though.)

I've identified potential cause of the issue:
 https://github.com/telefonicaid/iotagent-node-lib/blob/8a6cbd1a8d4023f239195f1aa1534c75a487f139/lib/services/ngsi/entities-NGSI-v1.js#L243-L249

Below fix is required (line# 246)
```
    if (
        'autoprovision' in typeInformation &&
        /* jshint -W101 */
        typeInformation.autoprovision !== undefined  //This should be !== operator and not ===
            ? typeInformation.autoprovision === true
            : config.getConfig().appendMode === true
    ) {
```

kandarp-n:
Am I supposed to raise a pull request instead (because I already found a fix)? I'm new to the open source world, so please advise :)

fgalan:
iotagent-node-lib/lib/services/ngsi/entities-NGSI-v1.js corresponds to an obsolete version of the API [soon to be removed](https://github.com/telefonicaid/iotagent-node-lib/pull/995).

Thus, could you configure your IOTA in NGSIv2 mode and test how it goes? Or if the same problem occurs in the NGSIv2 part (I guess iotagent-node-lib/lib/services/ngsi/entities-NGSI-v2.js) it would be great if you could send it a PR with the fix.

Thanks for your feedback!

AlvaroVega:
@kandarp-n it seems a typo and you find a fix. Just edit that file in github as proposed and create a new Pull Request then.

[...]

PR #1046, opened 2021-05-25
use explicitAttrs from gruoup/service conf

AlvaroVega:
Issues:
https://github.com/telefonicaid/iotagent-node-lib/issues/1045
https://github.com/telefonicaid/iotagent-node-lib/issues/1016

- For device created by autoprovision
- For a device without explicitAttrs but defined in their group

similar to group conf problems with expressionLanguage https://github.com/telefonicaid/iotagent-node-lib/pull/1029/


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Fix: use explicitAttrs defined in group is not defined at device level (#1046)
```
```suggestion
- Fix: use explicitAttrs defined in group is not defined at device level (#1045, #1016)
```

As I understand  both issues are addressed by this PR.

fgalan:
Retaining the merge a bit, just in case #1003 could have a similar fix, to be included in this PR.

Issue #1053, opened 2021-05-26
apply jexl over a non valid context, allowing null/undefined values


AlvaroVega:
The following code should be removed or at least not executed by processExpressionAttributes
https://github.com/telefonicaid/iotagent-node-lib/blob/3bbc3c2c817d5f9913b2f74cc90b3b3dfd4cc955/lib/plugins/jexlParser.js#L145-L159

Maybe just not calling contextAvailable from:

https://github.com/telefonicaid/iotagent-node-lib/blob/3bbc3c2c817d5f9913b2f74cc90b3b3dfd4cc955/lib/plugins/expressionParser.js#L239-L245

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1054

PR #1054, opened 2021-05-26
 simplify contextAvailable for jexl parser

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/issues/1053

### DIFF
--- a/lib/plugins/jexlParser.js
+++ b/lib/plugins/jexlParser.js
@@ -138,7 +138,7 @@ function contextAvailable(expression, context) {
     try {
         let validContext = true;
         jexl.evalSync(expression, context);
-        return validContext;
+        return true;
     } catch (e) {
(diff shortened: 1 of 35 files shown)


lib/plugins/jexlParser.js
```diff
@@ -136,32 +136,8 @@ function isTransform(identifier) {
...
-            if (element.value === 'type' || element.value === 'coordinates') {
-                validContext = true;
-            }
-            return validContext === false;
-        });
-        if (validContext) {
-            jexl.evalSync(expression, context);
-        }
+        jexl.evalSync(expression, context);
         return validContext;
```
fgalan, line 141: Maybe better?

```suggestion
        return true;
```

And remove the `let validContext = true;` above

fgalan (approved):
LGTM

PR #1061, opened 2021-06-03
Task/apply autropovision over entities ngsiv2 3

AlvaroVega:
autoprovision: optional boolean: If true, APPEND is used upon measure arrival (thus effectively allowing autoprovisioned devices). If false, UPDATE is used open measure arrival (thus effectively avoiding autoprovisioned devices). This field is optional, so if it omitted then the global IoTAgent appendModel configuration is used.

comes from https://github.com/telefonicaid/iotagent-node-lib/pull/1051 (issue #1003)

- [x] Update Tests

PR #1068, opened 2021-06-09
Update advanced-topics.md

mapedraza:
Added explanation for:
- autoprovision
- explicitAttrs
- appendMode

### DIFF
--- a/doc/advanced-topics.md
+++ b/doc/advanced-topics.md
@@ -70,7 +70,7 @@ values. If the latitude and longitude are received as separate measures, the
 ```
 
 For `attributes` and `static_attributes` which need to be formatted as GeoJSON values, three separate input formats are
-accepted. Provided the `type` is provisioned correctly, the `value` may be defined using any of the following formats:
+accepted. Provided the `type` is oned correctly, the `value` may be defined using any of the following formats:
 
 -   a comma delimited string
 
@@ -417,6 +417,27 @@ subscription payload.
 For each attribute in the `reverse` array, an expression must be defined to calculate its value based on the
 notification attributes. This value will be passed to the underlying protocol with the `object_id` name. Details about
 how the value is then progressed to the device are protocol-specific.
+
+#### Autoprovision configuration (autoprovision)
+
+By default, when a measure arrives to the IoTAgent, if the `device_id` does not match with an existing one, then, the IoTA 
+creates a new device and a new entity according to the group config. Defining the parameter `autoprovision` to `false` 
+when provisioning the device or the group, the IoTA to reject the measure at the southbound, allowing only to persist the 
+data to devices that are already provisioned. 
+
+#### Explicitly defined attributes (explicitAttrs)
+
+If the parameter is not defined in the device or group provision, this parameter is stored in the Context Broker by adding 
+a new attribute to the entity with the same name of the parameter. By adding the parameter `explicitAttrs` with `true` value
+to device or group provision, the IoTAgent rejects the parameters received from the measure that are not defined or in the 
+device or group provision, persisting only the one defined in the provision.
+
+#### Configuring operation to persist the data in Context Broker (appendMode)
+
+This is a flag that can be enabled by activating the parameter `appendMode` in the configuration file or by using the `IOTA_APPEND_MODE`
+environment variable	 (more info [here](https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/installationguide.md). 
+If this flag is activated, the update requests to the Context Broker will be performed always with APPEND type, instead of the 
+default UPDATE. This have implications in the use of attributes with Context Providers, so this flag should be used with care.
 
 ### Old IoTAgent data migration
 
--- a/doc/advanced-topics.md
+++ b/doc/advanced-topics.md
@@ -4,19 +4,23 @@
 * [GeoJSON support this only applies for NGSI LD, not for NGSIv1 and NGSIv2](#geojson-support-this-only-applies-for-ngsi-ld-not-for-ngsiv1-and-ngsiv2)
 * [Metadata support](#metadata-support)
     * [NGSI LD data and metadata considerations](#ngsi-ld-data-and-metadata-considerations)
+* [NGSI-LD Linked Data support](#ngsi-ld-linked-data-support)
+* [Autoprovision configuration (autoprovision)](#autoprovision-configuration-autoprovision)
+* [Explicitly defined attributes (explicitAttrs)](#explicitly-defined-attributes-explicitattrs)
+* [Configuring operation to persist the data in Context Broker (appendMode)](#configuring-operation-to-persist-the-data-in-context-broker-appendmode)
 * [Data mapping plugins](#data-mapping-plugins)
     * [Development](#development)
     * [Provided plugins](#provided-plugins)
-        * [Timestamp Compression plugin compressTimestamp](#timestamp-compression-plugin-compresstimestamp)
-        * [Attribute Alias plugin attributeAlias](#attribute-alias-plugin-attributealias)
-        * [Event plugin addEvents](#event-plugin-addevents)
-        * [Timestamp Processing Plugin timestampProcess](#timestamp-processing-plugin-timestampprocess)
-        * [Expression Translation plugin expressionTransformation](#expression-translation-plugin-expressiontransformation)
-        * [Multientity plugin multiEntity](#multientity-plugin-multientity)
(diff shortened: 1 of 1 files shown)


doc/advanced-topics.md
```diff
@@ -418,6 +418,27 @@ For each attribute in the `reverse` array, an expression must be defined to calc
 notification attributes. This value will be passed to the underlying protocol with the `object_id` name. Details about
 how the value is then progressed to the device are protocol-specific.
 
+#### Autoprovision configuration (autoprovision)
+
+By default, when a measure arrives to the IoTAgent, if the `device_id` does not match with an existing one, then, the IoTA 
+creates a new device and a new entity according to the group config. Defining the parameter `autoprovision` to `false` 
+when provisioning the device or the group, the IoTA to reject the measure at the southbound, allowing only to persist the 
+data to devices that are already provisioned. 
+
```
fgalan, line 427: I'd suggest to add the following paragraph:

> `If `autoprovision` is provided both at device and group level, the device level takes precedence

Does it work that way, right?

doc/advanced-topics.md
```diff
@@ -418,6 +418,27 @@ For each attribute in the `reverse` array, an expression must be defined to calc
...
+#### Explicitly defined attributes (explicitAttrs)
+
+If the parameter is not defined in the device or group provision, this parameter is stored in the Context Broker by adding 
+a new attribute to the entity with the same name of the parameter. By adding the parameter `explicitAttrs` with `true` value
+to device or group provision, the IoTAgent rejects the parameters received from the measure that are not defined or in the 
+device or group provision, persisting only the one defined in the provision.
+
```
fgalan, line 434: I'd suggest to add the following paragraph:

> `If `explicitAttrs` is provided both at device and group level, the device level takes precedence

Does it work that way, right?

doc/advanced-topics.md
```diff
@@ -418,6 +418,27 @@ For each attribute in the `reverse` array, an expression must be defined to calc
 notification attributes. This value will be passed to the underlying protocol with the `object_id` name. Details about
 how the value is then progressed to the device are protocol-specific.
 
+#### Autoprovision configuration (autoprovision)
```
fgalan, line 421: This PR adds three subsection at 4-level. The containing 3-level section is:

```
### Data mapping plugins
```

However, autoprovision/explicitAttr/appendMode are not relate with data mapping plugins.

Structure should be reviewed.

mapedraza: Should we create a new section? In which case I don't know if we should group the 3 new descriptions under the same section or not.

doc/advanced-topics.md
mapedraza: Just reading your review, I just realized It does not have sense to define autoprovision at device leve since this means the device is arlready provisioned.

mapedraza: Fixed in https://github.com/telefonicaid/iotagent-node-lib/pull/1068/commits/1c058365f8b3fd607d91def65af548f7c86a27d8

doc/advanced-topics.md
mapedraza: Added in https://github.com/telefonicaid/iotagent-node-lib/pull/1068/commits/1c058365f8b3fd607d91def65af548f7c86a27d8

doc/advanced-topics.md
fgalan: Not sure... I lack a global pespective of the topics in this file.

To know that, it would be great to do a multi-level ToC at the beginning of the file. Something like this: https://github.com/telefonicaid/fiware-cygnus/tree/master/doc/cygnus-ngsi/user_and_programmer_guide

doc/advanced-topics.md
fgalan: > Just reading your review, I just realized It does not have sense to define autoprovision at device leve since this means the device is arlready provisioned.

Good point :)

Maybe it should be even explicitly mentioned.

doc/advanced-topics.md
mapedraza: Added in https://github.com/telefonicaid/iotagent-node-lib/pull/1068/commits/9ecb677e15323c82b88a0b1dce2fc620287fa1ec

doc/advanced-topics.md
mapedraza: Added in https://github.com/telefonicaid/iotagent-node-lib/pull/1068/commits/9ecb677e15323c82b88a0b1dce2fc620287fa1ec

doc/advanced-topics.md
fgalan: After looking the ToC (thanks for put it together!) I don't see the new three topics fit in any existing group, so I'd suggest to add then at top-most level (maybe between "Metadata support" and "Data mapping plugins" topics).

doc/advanced-topics.md
```diff
@@ -1,5 +1,25 @@
 ## Advanced Topics
 
+* [Secured access to the Context Broker](#secured-access-to-the-context-broker)
+* [GeoJSON support this only applies for NGSI LD, not for NGSIv1 and NGSIv2](#geojson-support-this-only-applies-for-ngsi-ld-not-for-ngsiv1-and-ngsiv2)
```
fgalan, line 4: It seems linter doesn't like this line:

imagen

If changed, I understand the actual title in line 69 should be also be changed.

fgalan: Or maybe better "NGSI-v2"...

Not sure which is better ("NGSI v2" or "NGSI-v2") from a English language point of view. Check other uses in existing documentation.

mapedraza: In other parts in this repo is used both NGSI-v2 and NGSI v2, but seems more used the first one. On Orion repo is used NGSIv2.

doc/advanced-topics.md
mapedraza: Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1068/commits/99a7114e963e206a1b339f05f2bdf173cc7cef46

AlvaroVega (approved):
LGTM

fgalan (approved):
LGTM. Good documentation work!

I'd suggest to wait for https://github.com/telefonicaid/iotagent-node-lib/pull/1061 and merge both at the same time.

doc/advanced-topics.md
```diff
@@ -237,6 +261,29 @@ updated as shown:
 }
 ```
 
+### Autoprovision configuration (autoprovision)
+
+By default, when a measure arrives to the IoTAgent, if the `device_id` does not match with an existing one, then, the IoTA 
+creates a new device and a new entity according to the group config. Defining the parameter `autoprovision` to `false` 
+when provisioning the device group, the IoTA to reject the measure at the southbound, allowing only to persist the 
+data to devices that are already provisioned. It makes no sense to use this parameter in device provisioning since it is 
+intended to avoid provisioning devices (and for it to be effective, it would have to be provisional).
```
fgalan, line 270:
```suggestion
By default, when a measure arrives to the IoTAgent, if the `device_id` does not match with an existing one, then, the IoTA 
creates a new device and a new entity according to the group config. Defining the field `autoprovision` to `false` 
when provisioning the device group, the IoTA to reject the measure at the southbound, allowing only to persist the 
data to devices that are already provisioned. It makes no sense to use this field in device provisioning since it is 
intended to avoid provisioning devices (and for it to be effective, it would have to be provisional).
```

doc/advanced-topics.md
```diff
@@ -237,6 +261,29 @@ updated as shown:
...
+
+### Explicitly defined attributes (explicitAttrs)
+
+If the parameter is not defined in the device or group provision, this parameter is stored in the Context Broker by adding 
+a new attribute to the entity with the same name of the parameter. By adding the parameter `explicitAttrs` with `true` value
+to device or group provision, the IoTAgent rejects the parameters received from the measure that are not defined or in the 
+device or group provision, persisting only the one defined in the provision. If `explicitAttrs` is provided both at device 
+and group level, the device level takes precedence.
```
fgalan, line 278:
```suggestion
If a given measure is not defined in the mappings of the device or group provision, the measure is stored in the Context Broker by adding 
a new attribute to the entity with the same name of the measure. By adding the field `explicitAttrs` with `true` value
to device or group provision, the IoTAgent rejects the measures that are not defined in the 
mappings of device or group provision, persisting only the one defined in the mappings of the provision. If `explicitAttrs` is provided both at device 
and group level, the device level takes precedence.
```

fgalan (approved):
LGTM

PR #1069, opened 2021-06-11
fix typo in the example of expression legacy

danielvillalbamota:
Fix the example where attributes in the expression miss the `@` and remove the `cast` parameter in the provision as it returns an error 400:
```
{
    "name": "WRONG_SYNTAX",
    "message": "Wrong syntax in request: Errors found validating request."
}
```
It seems to be an old parameter, it is not described in the documentation, and the cast function it is provided by default using the provide "type" parameter


fgalan (approved):
LGTM. Thanks for the typo fixing!

PR #1070, opened 2021-06-15
Fix typo in bidirectional example documentation

danielvillalbamota:
Fix typo in the expression of the attributes that miss `@` as prefix.

Add a note about using `IOTA_AUTOCAST=false` to make the example works in the reverse side which expect the attribute as geo:point and no geo:json that is converted with `AUTOCAST=true` by default.


fgalan (approved):
LGTM

Issue #1076, opened 2021-07-05
Differences in provisioning with 2.16.0 compared to 2.15.2


jason-fox:
The following provisioning was made with **Ultralight**, but it shouldn't make a difference, since I make no measures or commands:

1. Provision two service groups

```console
curl -s -o /dev/null -X POST \
  'http://iot-agent:4041/iot/services' \
  -H 'Content-Type: application/json' \
  -H 'fiware-service: openiot' \
  -H 'fiware-servicepath: /' \
  -d '{
 "services": [
   {
     "apikey":      "321701236",
     "cbroker":     "http://orion:1026",
     "entity_type": "TemperatureSensor",
     "resource":    "/iot/d",
     "protocol":    "PDI-IoTA-UltraLight",
     "transport":   "HTTP",
     "timezone":    "Europe/Berlin",
     "attributes": [
       { "object_id": "t", "name":"temperature", "type": "Integer",
         "metadata": {"unitCode": {"type": "Property","value": "CEL"}}
       }
      ],
      "static_attributes": [
        {"name": "description", "type":"Property", "value": "Temperature Sensor"},
        {"name": "category", "type":"Property", "value": ["sensor"]},
        {"name": "controlledProperty", "type": "Property", "value": "temperature"},
        {"name": "supportedProtocol", "type": "Property", "value": ["ul20"]}
      ]
   },
   {
     "apikey":      "854782081",
     "cbroker":     "http://orion:1026",
     "entity_type": "FillingLevelSensor",
     "resource":    "/iot/d",
     "protocol":    "PDI-IoTA-UltraLight",
     "transport":   "HTTP",
     "timezone":    "Europe/Berlin",
     "attributes": [
       { "object_id": "f", "name":"filling", "type": "Number",
         "metadata": {"unitCode": {"type": "Property","value": "C62"}}
       }
      ],
      "commands": [
        {"name": "add", "type": "command"},
        {"name": "remove", "type": "command"}
      ],
      "static_attributes": [
        {"name": "supportedProtocol", "type": "Property", "value": ["ul20"]},
        {"name": "description", "type":"Property", "value": "Barn filling Sensor"},
        {"name": "category", "type":"Property", "value": ["actuator", "sensor"]},
        {"name": "controlledProperty", "type": "Property", "value": ["filling"]}
      ]
   }
 ]
}'
```

2. Provision two devices 

```console
curl -s -o /dev/null -X POST \
[...]

PR #1077, opened 2021-07-06
findByType and then findWihtoutType if no group is found

AlvaroVega:
issue reported https://github.com/telefonicaid/iotagent-node-lib/issues/1076
related with PR https://github.com/telefonicaid/iotagent-node-lib/pull/1010/

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,0 +1,1 @@
+- Fix: Try to search for a device group service in base of apikey and type (#1076)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,1 @@
-- Fix: Try to search for a device group service in base of apikey and type (#1076)
+- Fix: try to search for a device group service in base of not only by apikey but also type (#1076)
--- a/doc/api.md
+++ b/doc/api.md
@@ -203,6 +203,21 @@ outgoing requests).
 Note that there is a 1:1 correspondence between payload fields and DB fields (but using a different capitalization, e.g.
 `service_path` vs. `servicePath`).
 
+### Relationship between service groups and devices
+
+Devices may be associated to exisiting service groups (or not) based in `apiKey` matching or `type` matching (in the case `apiKey`
+matching fails). For instance, let's consider a situation in which a service group has been provisioned with `type=X`/`apiKey=111`
+and no other service group has been provisioned.
+
+* IoT Agent receives an anonymous measure with `apiKey=111`. The matching `apiKey` means the entity inherits from service group.
+  Device entity has `type=X` and `apiKey=111`
+* IoT Agent receives a provisioning request for an explicit device of `type=Y`/`apiKey=111`. The matching `apiKey` means the entity
+  inherits from service group but type is overridden. Device entity has `type=Y` and  `apiKey=111`
+* IoT Agent receives a provisioning request for an explicit device of `type=X`/`apiKey=222`. The matching `type` means the entity
+  inherits from service group but `apiKey` is overridden. Device entity has `type=X` and `apiKey=222`.
+* IoT Agent receives a provisioning request for  an explicit device of `type=Y`/`apiKey=222`. No matching. Device entity has `type=Y`
+  and `apiKey=222` and no service group.
+
 ### Device model
 
 The table below shows the information held in the Device resource. The table also contains the correspondence between
--- a/lib/services/devices/deviceService.js
+++ b/lib/services/devices/deviceService.js
@@ -190,6 +190,16 @@ function mergeDeviceWithConfiguration(fields, defaults, deviceData, configuratio
  * @param {Object} deviceObj        Device data.
  */
 function findConfigurationGroup(deviceObj, callback) {
+    function handlerGroupFindByType(error, group) {
+        if (!group) {
+            config
+                .getGroupRegistry()
+                .findSilently(deviceObj.service, deviceObj.subservice, deviceObj.apikey, handlerGroupFind);
+        } else {
+            handlerGroupFind(error, group);
+        }
+    }
+
     function handlerGroupFind(error, group) {
         let effectiveGroup = group;
 
@@ -207,7 +217,7 @@ function findConfigurationGroup(deviceObj, callback) {
     } else {
         config
             .getGroupRegistry()
-            .findSilently(deviceObj.service, deviceObj.subservice, deviceObj.apikey, handlerGroupFind);
+            .findTypeSilently(deviceObj.service, deviceObj.subservice, deviceObj.type, deviceObj.apikey, handlerGroupFind);
     }
 }
 
--- a/lib/services/devices/deviceService.js
+++ b/lib/services/devices/deviceService.js
@@ -217,7 +217,7 @@ function findConfigurationGroup(deviceObj, callback) {
     } else {
         config
             .getGroupRegistry()
-            .findTypeSilently(deviceObj.service, deviceObj.subservice, deviceObj.type, deviceObj.apikey, handlerGroupFind);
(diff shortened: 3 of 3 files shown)


jason-fox:
1. Send the following Service Group provisioning:

```json
{
 "services": [
  {
     "apikey":      "110990",
     "entity_type": "Device",
     "resource":    "/iot/d",
     "attributes": [
        {"object_id": "bpm","type": "Integer","name": "heartRate",
        "metadata": {"unitCode": {"type": "Text","value": "5K"}}
        },
        {"object_id": "s", "name": "status", "type": "String"},
        {"object_id": "gps", "name": "location", "type": "geo:point"}
      ],
      "static_attributes": [
        {"name": "description", "type":"Property", "value": "Animal Collar"},
        {"name": "category", "type":"Property", "value": ["sensor"]},
        {"name": "controlledProperty", "type": "Property", "value": ["heartRate", "location", "status"]}
      ]
   }
   {
     "apikey":      "321701236",
     "entity_type": "TemperatureSensor",
     "resource":    "/iot/d",
     "attributes": [
       { "object_id": "t", "name":"temperature", "type": "Integer",
         "metadata": {"unitCode": {"type": "Property","value": "CEL"}}
       }
      ],
      "static_attributes": [
        {"name": "description", "type":"Property", "value": "Temperature Sensor"},
        {"name": "category", "type":"Property", "value": ["sensor"]},
        {"name": "controlledProperty", "type": "Property", "value": "temperature"}
      ]
   },
   {
     "apikey":      "854782081",
     "entity_type": "FillingLevelSensor",
     "resource":    "/iot/d",
     "attributes": [
       { "object_id": "f", "name":"filling", "type": "Number",
         "metadata": {"unitCode": {"type": "Property","value": "C62"}}
       }
      ],
      "commands": [
        {"name": "add", "type": "command"},
        {"name": "remove", "type": "command"}
      ],
      "static_attributes": [
        {"name": "supportedProtocol", "type": "Property", "value": ["ul20"]},
        {"name": "description", "type":"Property", "value": "Barn filling Sensor"},
        {"name": "category", "type":"Property", "value": ["actuator", "sensor"]}
      ]
   }
 ]
}
```

2. Send an anonymous measure to `/?id=collar001&apiKey=110990`

3. Now explicitly provision three devices:

```console
{
 "devices": [
    {
       "device_id":   "humidity001",
       "entity_name": "urn:ngsi-ld:SoilHumidity:humidity001",
       "entity_type": "SoilHumidity",
       "static_attributes": [
           {"name": "controlledAsset", "type": "Relationship","value": "urn:ngsi-ld:Building:farm001"}
       ]
    },
    {
       "device_id":   "temperature001",
       "entity_name": "urn:ngsi-ld:TemperatureSensor:temperature001",
       "entity_type": "TemperatureSensor",
       "static_attributes": [
           {"name": "controlledAsset", "type": "Relationship","value": "urn:ngsi-ld:Building:farm001"}
       ]
    },
    {
      "device_id": "filling001",
      "entity_name": "urn:ngsi-ld:FillingLevelSensor:filling001",
      "entity_type": "FillingLevelSensor",
      "endpoint": "http://context-provider:3001/iot/filling001",
      "static_attributes": [
        {"name": "controlledAsset", "type": "Relationship","value": "urn:ngsi-ld:Building:farm001"}
      ]
    }
  ]
}
'
```

I would expect 4 devices in the context broker.

1. The Anonymous device inherits from `Device` through the `apiKey`, `urn:ngsi-ld:Device:collar001` has `heartRate` and `location`
2. `urn:ngsi-ld:TemperatureSensor:temperature001` inherits from the `TemperatureSensor` service group and has `temperature` 
3. `urn:ngsi-ld:FillingLevelSensor:filling001` inherits from the `FillingLevelSensor` service group and has `add` and `remove` commands registered 
4. What should `urn:ngsi-ld:SoilHumidity:humidity001` hold?

jason-fox:
One idea would be to add the following fallbacks:

1. Add a new Docker `ENV` / `config.js` param to define the `type` to inherit from if no explicit `type` found.
2. If nothing found (or the override is not set?)  just pick the first one ???

The issue here is that the `SoilSensor` could conceivably inherit from any of the provisioned services.

AlvaroVega:
fixed in https://github.com/telefonicaid/iotagent-node-lib/pull/1077/commits/f6c1b375477eab69d16f1dfff97e8b5d2a6fbb4f could you retry?

jason-fox:
My original issue is fixed, but the following is still unclear:

1. I provision two service Groups `A` and `B`
2.  I provision a device with entity type `C` - what does it inherit?

On my test run it inherited from the **first**  provisioned service group, but this needs to be made clear in the documentation. Alternatively provide a workaround (e.g. fail if multiple groups available?? or add a default in the `config.js` or something.) My preference is for a fallback to an explicit default (e.g. `Thing` in the `config.js`) and if that doesn't exist pick the first.

Without `type=XXX` in the database query, I don't think mongoDB will necessarily  choose the same default each time, so the result of the provisioning (in this case) is not clear - at least if the docs said something it would help.

AlvaroVega:
> My original issue is fixed, but the following is still unclear:
> 
>     1. I provision two service Groups `A` and `B`
> 
>     2. I provision a device with entity type `C` - what does it inherit?
> 
> 
> On my test run it inherited from the **first** provisioned service group, but this needs to be made clear in the documentation. Alternatively provide a workaround (e.g. fail if multiple groups available?? or add a default in the `config.js` or something.) My preference is for a fallback to an explicit default (e.g. `Thing` in the `config.js`) and if that doesn't exist pick the first.
> 
> Without `type=XXX` in the database query, I don't think mongoDB will necessarily choose the same default each time, so the result of the provisioning (in this case) is not clear - at least if the docs said something it would help.

If A is group type "A" and B is a group type "B" and a device C is a type "C" and apikey of C if provided is not from A or B then C could not inherit nothing from A or B groups

jason-fox:
> There is some information about group services in https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/api.md#service-group-api

I guess the documentation needs expanding here to clarify the various scenarios - something like the text below. It always helps to have an explicit example:


When sending NGSI Entity requests  to the context broker, the payload will include information from the provisioned service group if the `entity_type` and/or the `apiKey` matches. 

For example if a  service group  is provisioned with `type=X`/`apiKey=111` and other attributes then the following will occur:

-   IoT Agent receives an anonymous measure with `apiKey=111`
    inherits from service group. Device entity has `type=X` and  `apiKey=111`
-   IoT Agent receives a provisioning request for an explicit device of `type=Y`/`apiKey=111`
   The matching `apiKey` means the entity inherits from service group but type is overridden. Device entity has `type=Y` and  `apiKey=111`
-   IoT Agent receives a provisioning request for  an explicit device of `type=X`/`apiKey=222`
   The matching `type` means the entity inherits from service group but `apiKey` is overridden. Device entity has `type=x` and  `apiKey=222`
-  IoT Agent receives a provisioning request for  an explicit device of `type=Y`/`apiKey=222`
   No matching. Device entity has `type=Y` and  `apiKey=222`

fgalan:
Thank you for your suggestions @jason-fox . Based on it, we have improved documentation in f98b5b9, adding a section in api.md to explain this better. If you think is there any other better place to put that section or have any other suggestion, feedback is welcome :)

However, note that the case:

> * IoT Agent receives a provisioning request for  an explicit device of `type=Y`/`apiKey=222`. No matching. Device entity has `type=Y` and `apiKey=222` and no service group.

contradicts what you say about:

> On my test run it [a device which doesn't match] inherited from the first provisioned service group

Or maybe I'm missing something?

CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Fix: Try to search for a device group service in base of apikey and type (#1076)
```
```suggestion
- Fix: try to search for a device group service in base of not only by apikey but also type (#1076)
```

fgalan (approved):
LGTM

AlvaroVega (commented):
LGTM

Issue #576, opened 2018-02-01
Bidirectional plugin doesn't perform unsubscription on device removal


fgalan:
The library has methods that implementes the unsubscribe operation on CB. In particular, the unsubscribe() function at https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/ngsi/subscriptionService.js#L184

```
function unsubscribe(device, id, callback) {
    ...
    if (device.cbHost) {
        options.uri = 'http://' + device.cbHost + '/v1/unsubscribeContext';
    } else {
        options.uri = 'http://' + config.getConfig().contextBroker.host + ':' + config.getConfig().contextBroker.port +
            '/v1/unsubscribeContext';
    }
...
}
```
which is called by removeAllSubscriptions() function at https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/services/devices/deviceService.js#L415:

```
function removeAllSubscriptions(device, callback) {
    function removeSubscription(subscription, callback) {
        subscriptions.unsubscribe(device, subscription.id, callback);
    }

    if (device.subscriptions) {
        async.map(device.subscriptions, removeSubscription, callback);
    } else {
        callback(null, {});
    }
}
```
However, nobody in the library is using this method, as can be checked by this search:

* https://github.com/telefonicaid/iotagent-node-lib/search?utf8=%E2%9C%93&q=removeAllSubscriptions&type= (0 hits)

In addition, IOTA agents hasn't any link with this code in the library, I mean nobody using it the IOTA JSON agent:

*https://github.com/telefonicaid/iotagent-json/search?utf8=%E2%9C%93&q=removeAllSubscriptions&type= (0 hits)
* https://github.com/telefonicaid/iotagent-json/search?utf8=%E2%9C%93&q=unsubscribe&type= (1 hits in no test code but related with MQTT)

and nobody using in the IOTA UL agent:

* https://github.com/telefonicaid/iotagent-ul/search?utf8=%E2%9C%93&q=removeAllSubscriptions&type= (0 hits)
* https://github.com/telefonicaid/iotagent-ul/search?utf8=%E2%9C%93&q=unsubscribe&type= (1 hit in no test code but related with MQTT)

If my theory confirms (I may be wrong) it seems that unsubcription code in the library is dead code. As a consecuence, subscriptions are not correctly cleaned up when devices are deleted by API, leaving dirty stuff in CB.

chicco785:
[...]

PR #1082, opened 2021-08-20
remove the removeAllSubscriptions function

Sunny-NEC:
pr for the issue #576

### DIFF
--- a/lib/services/devices/deviceService.js
+++ b/lib/services/devices/deviceService.js
@@ -354,17 +354,6 @@ function registerDevice(deviceObj, callback) {
     );
 }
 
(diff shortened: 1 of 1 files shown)


Sunny-NEC:
image

@fgalan I did some test but did not find any issue after removing this function.

fgalan:
> @fgalan I did some test but did not find any issue after removing this function.

Yes, but that's not the way of solving issue #576 :) The good solution is to properly use that code, not removing it. Please find my detailed comment [above](https://github.com/telefonicaid/iotagent-node-lib/pull/1082#issuecomment-906161732).

Issue #1056, opened 2021-05-26
allow load jexl transformations from config


AlvaroVega:
config.js

var transformations = [ ];

in jexlParser

  jexl.addTransformations((config.transformations);

Madhu-NEC:
Hello @AlvaroVega ,

As per my understanding, for this issue we need to perform jexl transformations from config, if provided, otherwise the same is loaded as mentioned [here](https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/plugins/jexlParser.js#L38)

If my understanding is not correct, please share your opinion and elaborate on this issue.

fgalan:
> As per my understanding, for this issue we need to perform jexl transformations from config, if provided, otherwise the same is loaded as mentioned [here](https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/plugins/jexlParser.js#L38)

Not fully sure but what I understand is that this issue is about *adding* (not *replacing*) custom user functions (defined in config.js file) to the ones in the fragment of code you mention.

How to deal with conflict? For instance, we have `indexOf` defined in jexlParser.js. What would happen if the user defines a function with the same name in config.js? A preference rule should be defined and documented.

Sunny-NEC:
@fgalan  @AlvaroVega According to my investigation, I have found that the setting of configuration from config.js is done here(https://github.com/telefonicaid/iotagent-json/blob/master/lib/iotagent-json.js#L116). But jexl.addTransform() functions of jexlParser.js are executed before the setting of configuration. 
Above execution of code restricts the execution of jexlTransform() after the setting of configuration.

To resolve the above scenario, I have below approach:
I am trying to put all the jexlTransform() statements into a new function and then call the new function here(https://github.com/telefonicaid/iotagent-json/blob/master/lib/iotagent-json.js#L123) so that the jexlTransform() executes after setting the configuration and before executing any dataPlugins.

Please share your opinion on above approach.

fgalan:
Reading the comment I'm not sure about your approach. It would be better if you could create a pull request to show it... best way of see this king of things in on code :)

Thanks for your contribution!

Sunny-NEC:
Thanks for the response @fgalan 

Thus, if the user has this in the config.js file:

    transformations : [   { transformation_Name : 'index', transformation_Body : '(val, char) => String(val).indexOf(char)'} , 
                    { transformation_Name : 'trim', transformation_Body : '(val) => String(val).trim()'}
[...]

PR #1090, opened 2021-08-24
Extract transform definiton to a config file #1056

mrutid:
Just extract JEXL Transforms to a top level module.

### DIFF
--- a/lib/plugins/jexlParser.js
+++ b/lib/plugins/jexlParser.js
@@ -31,45 +31,12 @@ const jexl = require('jexl');
 const errors = require('../errors');
 const logger = require('logops');
 const config = require('../commonConfig');
+const tranformsMap = require('../jexlTranformsMap.js').map;
 const logContext = {
     op: 'IoTAgentNGSI.JEXL'
 };
-jexl.addTransform('jsonparse', (str) => JSON.parse(str));
-jexl.addTransform('jsonstringify', (obj) => JSON.stringify(obj));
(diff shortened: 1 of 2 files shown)


lib/plugins/jexlParser.js
```diff
@@ -185,8 +150,53 @@ function processExpressionAttributes(typeInformation, list, context) {
         .map(expressionApplier(context, typeInformation));
 }
 
+function checkTransformationMap(tranformsMap) {
+    //[error,message, resultMap] = expressionParser.checkTransformationMap(niceMap);
```
fgalan, line 154: Leftover comment? Should be removed?

lib/plugins/jexlParser.js
```diff
@@ -185,8 +150,53 @@ function processExpressionAttributes(typeInformation, list, context) {
...
+            message = 'Trasformations can be added to JEXL parser';
+        } else {
+            message = wrongList.toString() + ' must be a function';
+        }
+    }
+    return [error, message, resultMap];
+}
+
+function setTransforms(configMap) {
+    //Check provided transforms since they are provided by user
```
fgalan, line 186: Maybe is a good idea to document what happen when name conflict occurs, something like this:

```suggestion
    //Check provided transforms since they are provided by user
    //Note that in case of name conflict, baseTransformsMap would take precedence
```

lib/plugins/jexlParser.js
mrutid: fixed

Issue #1109, opened 2019-09-04
Wrong http error code when provisioning a device with wrong static attributes


manucarrace:
Trying to create a device through manager with wrong location (cordinates between "" are not valid for Context broker)

{
    "devices": [
        {
            "device_id": "PA-VILL-ALU-005",
            "entity_name": "PA-VILL-ALU-005",
            "entity_type": "StreetlightControlCabinet",
            "static_attributes": [
                {
                    "name": "location",
                    "type": "geo:json",
                    "value": {
                        "type": "Point",
                        "coordinates": [
                            "-0.038202",
                            "39.976196"
                        ]
                    }
                }
            ],
            "protocol": "IoTA-JSON"
        }
    ]
}

The manager responds with a 200 OK, but the device is not created.

The iota makes the following request/response to/from the context broker:

POST /v2/entities?options=upsert HTTP/1.1
fiware-service: demo
fiware-servicepath: /castellon
fiware-correlator: 4144375b-bd44-4f56-9c4d-3d1ef0d50826
host: 10.0.0.2:1026
accept: application/json
content-type: application/json
content-length: 224
Connection: close

{"id":"PA-VILL-ALU-005","type":"StreetlightControlCabinet","location":{"type":"geo:json","value":{"type":"Point","coordinates":["-0.038202","39.976196"]}},"TimeInstant":{"type":"DateTime","value":"2019-09-04T06:18:25.315Z"}}HTTP/1.1 500 Internal Server Error
Server: nginx
Date: Wed, 04 Sep 2019 06:18:25 GMT
Content-Type: application/json
Content-Length: 1948
Connection: close
Fiware-Correlator: 4144375b-bd44-4f56-9c4d-3d1ef0d50826

[...]

PR #1110, opened 2021-09-24
set status code response from CB for EntityGenericError

AlvaroVega:
fix for issue https://github.com/telefonicaid/iotagent-node-lib/issues/1109

Previous this PR iota was responding 200

A EntityGenericError without explicit status code implires iotagent respond with 200 code:
https://github.com/telefonicaid/iotagent-node-lib/blob/e85304dd257addcfac0f9ceaddf8d1e9ddf82463/lib/errors.js#L38-L45

Maybe log error message "Protocol error connecting to the Context Broker" is also not right


AlvaroVega:
At this point is also returning a EntityGenericError without explicit status code, when CB has been returned 404 error after a query/update:
https://github.com/telefonicaid/iotagent-node-lib/blob/e85304dd257addcfac0f9ceaddf8d1e9ddf82463/lib/services/ngsi/entities-NGSI-v2.js#L228-L246

fgalan (approved):
LGTM

Issue #956, opened 2020-12-02
analize/ensure ngsiv2 + options=upsert is used to create/update entities


AlvaroVega:

AlvaroVega:
Upsert is used to create devices:
https://github.com/telefonicaid/iotagent-node-lib/blob/e85304dd257addcfac0f9ceaddf8d1e9ddf82463/lib/services/devices/devices-NGSI-v2.js#L213-L226

AlvaroVega:
But is not using upsert when update an entity:
https://github.com/telefonicaid/iotagent-node-lib/blob/e85304dd257addcfac0f9ceaddf8d1e9ddf82463/lib/services/devices/devices-NGSI-v2.js#L270-L280

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1111 (and associated PRs in IOTA repos: https://github.com/telefonicaid/iotagent-ul/pull/615 and https://github.com/telefonicaid/iotagent-json/pull/732)

PR #1111, opened 2021-09-24
use upsert when update entities

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/956

This will be enable to not create initially entities (actually is done even for autorpovisioned devices)


lib/services/devices/devices-NGSI-v2.js
```diff
@@ -269,7 +269,7 @@ function createInitialEntityNgsi2(deviceData, newDevice, callback) {
  */
 function updateEntityNgsi2(deviceData, updatedDevice, callback) {
```
fgalan, line 270: If the same CB API method is used at the end for both create and update cases, why do we need createInitialEntityNgsi2() and updateEntityNgsi2() separate functions? Maybe the can be unified? Or is some additional logic beyond CB request specific in each case?

AlvaroVega: Now both functions are very similar, but it seems they are not following the same logic

fgalan: In this PR on in a later one?

AlvaroVega: IMHO in another PR

fgalan: NTC

fgalan (approved):
LGTM

AlvaroVega:
Currently autorpovision is create initial entity and then performs an update. If we add this PR we can achieve in the future that autorpovision just update an entity without create it initially (see https://github.com/telefonicaid/iotagent-node-lib/issues/1334)

fgalan:
> Currently autorpovision is create initial entity and then performs an update. If we add this PR we can achieve in the future that autorpovision just update an entity without create it initially (see #1334)

@manucarrace what do you think?

fgalan:
This PR would make `appendMode` configuration setting useless. It would be as hardwiring this setting to always `true`. We would gain in simplicity (just one way to interact with CB) but we would loss flexibility.

AlvaroVega:
@fgalan how is related upsert with  appendMode of CB update?

https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/installationguide.md

appendMode: if this flag is activated, the update requests to the Context Broker will be performed always with APPEND type, instead of the default UPDATE. This have implications in the use of attributes with Context Providers, so this flag should be used with care. This flag is overwritten by autoprovision flag in group or device provision.

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/blob/ad90059d58a6c85b282547812dd0cd7ee01b97b6/lib/services/ngsi/entities-NGSI-v2.js#L351-L361

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -970,16 +970,19 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
...
             // recreate options object to use single entity update
-            url = '/v2/entities/' + entityName + '/attrs';
-            if (typeInformation && typeInformation.type) {
-                url += '?type=' + typeInformation.type;
+            url = '/v2/entities?options=upsert';
+            if (config.getConfig().appendMode === false) {
+                url = '/v2/entities/' + entityName + '/attrs';
+                if (typeInformation && typeInformation.type) {
+                    url += '?type=' + typeInformation.type;
+                }
```
fgalan, line 978:
```suggestion
            url = '/v2/entities';
            if (config.getConfig().appendMode === false) {
                if (typeInformation && typeInformation.type) {
                    url += '?type=' + typeInformation.type;
                }
            }
            else {  // false
                url += '?options=upsert
            }
```

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -970,16 +970,22 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
...
         if (options.json.entities.length === 1) {
             // recreate options object to use single entity update
-            url = '/v2/entities/' + entityName + '/attrs';
-            if (typeInformation && typeInformation.type) {
-                url += '?type=' + typeInformation.type;
+            url = '/v2/entities';
+            if (config.getConfig().appendMode === false) {
+                url += '/' + entityName + '/attrs';
+                if (typeInformation && typeInformation.type) {
+                    url += +'?type=' + typeInformation.type;
```
fgalan, line 977: I think this should be:

```suggestion
                    url += options.json.id +'/attrs?type=' + typeInformation.type;
```

(Not sure if `options.json.id` is the right variable here to get entity id)

lib/services/devices/devices-NGSI-v2.js
```diff
@@ -265,7 +265,7 @@ function createInitialEntityNgsi2(deviceData, newDevice, callback) {
...
     const options = {
-        url: config.getConfig().contextBroker.url + '/v2/entities/' + String(deviceData.name) + '/attrs',
+        url: config.getConfig().contextBroker.url + '/v2/entities/' + String(deviceData.name) + '/attrs?options=upsert',
```
fgalan, line 268: This operation is not going to work... CB only accepts `?options=upsert` in `POST /v2/entities` not in `POST /v2/entities/{id}/attrs`...

In addition, maybe `if (config.getConfig().appendMode === ...)` should be used also here?

lib/services/devices/devices-NGSI-v2.js
```diff
@@ -276,13 +276,14 @@ function updateEntityNgsi2(deviceData, updatedDevice, callback) {
     };
 
     if (deviceData.cbHost && deviceData.cbHost.indexOf('://') !== -1) {
-        options.url = deviceData.cbHost + '/v2/entities/' + String(deviceData.name) + '/attrs';
+        options.url = deviceData.cbHost + '/v2/entities/' + String(deviceData.name) + '/attrs?options=upsert';
     } else if (deviceData.cbHost && deviceData.cbHost.indexOf('://') === -1) {
-        options.url = 'http://' + deviceData.cbHost + '/v2/entities/' + String(deviceData.name) + '/attrs';
+        options.url =
+            'http://' + deviceData.cbHost + '/v2/entities/' + String(deviceData.name) + '/attrs?options=upsert';
     }
```
fgalan, line 283: Similar comment to the one above.

lib/services/devices/devices-NGSI-v2.js
AlvaroVega: Then I'll remove `?options=upsert` for that case. That case ins not the usual flow. An check apppendMode is not expected here

AlvaroVega: That flow is for update a register, and has not being modified; this PR just use upsert when update an entity after get a measure

CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
+- Fix: use 'options=upsert' when update ngsiv2 CB entities (#956)
```
```suggestion
- Fix: use 'options=upsert' when update ngsiv2 CB entities and appendMode is enabled (#956)
```

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -970,16 +970,22 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
...
-            if (typeInformation && typeInformation.type) {
-                url += '?type=' + typeInformation.type;
+            url = '/v2/entities';
+            if (config.getConfig().appendMode === false) {
+                url += '/' + entityName + '/attrs';
+                if (typeInformation && typeInformation.type) {
+                    url += '?type=' + typeInformation.type;
+                }
+            } else {
+                // false
```
fgalan, line 980:
```suggestion
                // appendMode === true
```

test/unit/ngsiv2/ngsiService/active-devices-test.js
```diff
@@ -193,26 +192,27 @@ describe('NGSI-v2 - Active attributes test', function () {
...
+        //                 value: 'wrongid'
+        //             }
+        //         ];
+
+        //         iotAgentLib.update('light1', 'Light', '', wrongvalues.concat(values), function (error) {
+        //             should.not.exist(error);
+        //             contextBrokerMock.done();
+        //             done();
+        //         });
+        //     });
```
fgalan, line 215: Should be re-enabled?

test/unit/ngsiv2/expressions/jexlBasedTransformations-test.js
```diff
@@ -1123,12 +1107,12 @@ describe('Java expression language (JEXL) based transformations plugin', functio
...
                 .matchHeader('fiware-servicepath', 'gardens')
                 .post(
-                    '/v2/entities/gps1/attrs',
+                    '/v2/entities?options=upsert',
                     utils.readExampleFile(
                         './test/unit/ngsiv2/examples/contextRequests/updateContextExpressionPlugin32.json'
                     )
                 )
-                .query({ type: 'GPS' })
+                //uery({ type: 'GPS' })
```
fgalan, line 1115: Lefotver?

AlvaroVega: Fixed in https://github.com/telefonicaid/iotagent-node-lib/pull/1111/commits/2fb5b110a37aacba924480c621934963462df562

fgalan (approved):
LGTM

Pendiong on @mapedraza review and LGTM

mapedraza (approved):
LGTM

PR #1116, opened 2021-10-07
[WIP] Update commandRegistryMongoDB.js

AlvaroVega:


AlvaroVega:
Actually we using `getCmd=1`

We need to provide a way of obtain ?
- All
- First
- Last

and include a way to empty commands

PR #1118, opened 2021-10-13
Amend null processing for NGSI-LD

jason-fox:
If an NGSI-LD  value is sent with a null attribute, remove it  from the NGSI-LD payload.

-    Update NGSI-LD entity processing
-    Update unit tests


jason-fox:
Related to https://github.com/FIWARE/context.Orion-LD/issues/919


For Orion-LD and the following workaround:

```json
"value": {
    "@type": "Intangible",
    "@value": null
}
```

It is causing inconsistency between brokers.  So it looks like this phrase will be tightened up, so the existing  _resolved to a`null` without being a `null`_ workaround will be made explicitly invalid by amending section 5.2.1


> *5.2.1	Introduction*
>
> Implementations shall support the data types defined by the clauses below. For each member defined by each data type 
> (including nested ones) a term shall be added to the Core `@context`, as mandated by clause 4.5.
> None of the members described admit a `null` value


I suggested to @Neeraj-Nec and @naveennec that a PR could be raised to deal with this. 

The fix is simple - if a value is parsed as `null` just set the attribute to `undefined` so it isn't sent.

PR #1135, opened 2021-11-10
Fix/multientity same name

AlvaroVega:
FIX simple case of multientity with same entity_names and different types.

Some problems maybe found with same measure or attribute names. More cases will follow


AlvaroVega (commented):
LGTM

PR #1130, opened 2021-11-05
Fix/wrong id type use as attrs #1124 

AlvaroVega:
clean PR version of https://github.com/telefonicaid/iotagent-node-lib/pull/1125

issue https://github.com/telefonicaid/iotagent-node-lib/issues/1124

Issue #1124, opened 2021-10-28
IOTA sends illegal entity type to CB (may also happend with entity id)


fgalan:
We have found that under some circustances (@danielvillalbamota please add the detail to this issue) IOTA sends this `POST /v2/op/update` to CB, with an illegal `type`

```
{
	"actionType": "append",
	"entities": [{
		"id": "GW02",
		"type": {
			"value": "2021-10-28T09:25:23.000",
			"type": "string"
		},
		"temperature": {
			"type": "Number",
			"value": "20",
			"metadata": {
				"TimeInstant": {
					"type": "DateTime",
					"value": "2021-10-28T07:57:06.706Z"
				}
			}
		},
		"TimeInstant": {
			"type": "DateTime",
			"value": "2021-10-28T07:57:06.706Z"
		}
	}, {
		"id": "HN01",
		"type": "heritageNode",
		"dateObserved": {
			"type": "DateTime",
			"value": "dateObs.type",
			"metadata": {
				"TimeInstant": {
					"type": "DateTime",
					"value": "2021-10-28T07:57:06.706Z"
				}
			}
		},
		"TimeInstant": {
			"type": "DateTime",
			"value": "2021-10-28T07:57:06.706Z"
		}
	}]
}
```

IOTA should avoid this, not only for `type` but also for `id`

fgalan:
Fixed by PR #1130 

Should this issue be closed? Or should be test before by @danielvillalbamota in the environment where originally happened?

fgalan:
> Fixed by PR #1130

I has been revered by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1136

PR #1139, opened 2021-11-10
Fix/wrong id type use as attrs #1124

AlvaroVega:
this PR is the same than https://github.com/telefonicaid/iotagent-node-lib/pull/1130 (that was reverted) but after merge https://github.com/telefonicaid/iotagent-node-lib/pull/1135


fgalan (approved):
LGTM

PR #1128, opened 2021-11-04
FIX simple case of multientity with same entity_names and different t…

mrutid:
FIX simple case of multientity with same entity_names and different types.

Some problems maybe found with same measure or attribute names. More cases will follow

PR #1141, opened 2021-11-11
revert simplification of entities

AlvaroVega:
The improvement was introduced in https://github.com/telefonicaid/iotagent-node-lib/pull/1128/commits/875e46c1069c1d206d5d958d880177a8fdf692b1

But in order keep entity in an array of entties iotagent-json and iotagent-ul tests cases (so many) should be updated.


fgalan (approved):
LGTM

mrutid (commented):
LGTM

Issue #1146, opened 2021-11-16
Review all commands to set/get log levels in IOTAs


cesarjorgemartinez:
Review all commands to set/get log levels in IOTAs

One attempt to set log level to WARN:
```
curl -X PUT http://host-iota-json:4052/admin/log?level=WARN
{"error":"invalid log level"}
```

Forces to use non standard style:
```
curl -X PUT http://host-iota-json:4052/admin/log?level=WARNING
```

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1147

cesarjorgemartinez:
The IOTA_LOG_LEVEL variable works correctly.
Testing launching `curl -X PUT http://host-iota-json:4052/admin/log?level=WARNING`:
The logs appeared as DEBUG.

AlvaroVega:
> The IOTA_LOG_LEVEL variable works correctly. Testing launching `curl -X PUT http://host-iota-json:4052/admin/log?level=WARNING`: The logs appeared as DEBUG.

In fact change to WARNING level was not working, since logs were still appearing with previous level

cesarjorgemartinez:
When set WARNING not back into the previous level (was in INFO). Directly changes to DEBUG.

PR #1147, opened 2021-11-17
Fix/suppong warn log level

AlvaroVega:
logops dep only wupport WARN log level
We need to provide backward compatibility for WARNING log level

https://github.com/telefonicaid/iotagent-node-lib/issues/1146


fgalan (approved):
LGTM

Issue #1122, opened 2021-10-22
Question: Should the library expose more common functions?


jason-fox:
#1119 has abstracted  HTTP requests from the IoT Agents, should other common library  functions such as `logops` be exposed by the IoT Agent node lib. Looking at the dependencies:

### iotagent-node-lib 

```json
"dependencies": {
    "async": "2.6.2",
    "body-parser": "~1.19.0",
    "command-shell-lib": "1.0.0",
    "express": "~4.16.4",
    "got": "~11.8.2",
    "jexl": "2.3.0",
    "jison": "0.4.18",
    "logops": "2.1.0",
    "moment": "~2.24.0",
    "moment-timezone": "~0.5.25",
    "mongodb": "3.6.8",
    "mongoose": "5.7.7",
    "mu2": "~0.5.20",
    "query-string": "6.5.0",
    "revalidator": "~0.3.1",
    "underscore": "~1.12.1",
    "uuid": "~3.3.2"
  }
```

### A random IoT Agent

```json
  "dependencies": {
    "amqplib": "~0.5.1",
    "async": "2.6.1",
    "body-parser": "1.18.3",
    "dateformat": "3.0.3",
    "express": "~4.16.4",
    "logops": "2.1.0",
    "mqtt": "3.0.0",
    "underscore": "1.12.1"
  }
```

Things like `async` or `mongoose` or `logops` or `express` or `underscore` could be exposed and standardized more easily across agents.

Also, do you really need both `dateformat` and `moment` ? Surely they are doing the same job?

Issue #1168, opened 2021-12-22
Refactor - remove underscore dependency from codebase


mapedraza:
The idea is to reduce the amount of dependencies used in codebase, specially for those which are not essential, in order to reduce possible vulnerabilities like Log4j CVE-2021-44228.

In the particular case of underscore, we have found that it is used only in these cases:
- pluck method
- difference and keys method used once [here](https://github.com/telefonicaid/iotagent-node-lib/blob/7748463ae2c709c088d35ed037d50bf733e88df4/lib/services/ngsi/entities-NGSI-v2.js#L336)

Both cases can be easily covered without using underscore (in the case of plunk, using a map filter, in the case of difference and key amending the logic in charge of that part of the code)

jason-fox:
@mapedraza  if you look at the `package-lock.json`, you'll notice that the `async` dependency relies on `lodash`:

```json
 "async": {
      "version": "2.6.2",
      "resolved": "https://registry.npmjs.org/async/-/async-2.6.2.tgz",
      "integrity": "sha512-H1qVYh1MYhEEFLsP97cVKqCGo7KfCyTt6uEWqsTBr9SO84oK9Uwbyd/yCW+6rKJLHksBNUVWZDAjfS+Ccx0Bbg==",
      "requires": {
        "lodash": "^4.17.11"
      }
    },
```

`lodash` is [nearly but not quite](https://www.geeksforgeeks.org/difference-between-lodash-and-underscore/) a one-to-one drop in for `underscore` - you could just switch the library for using `pluck()` and `map()`. This would eliminate `underscore` entirely since no other dependency is using it. It should improve throughput slightly too since `lodash` is supposedly more performant.

AlvaroVega:
lib/plugins/bidirectionalData.js:const _ = require('underscore');
lib/plugins/expressionParser.js:const _ = require('underscore');
lib/services/ngsi/entities-NGSI-LD.js:const _ = require('underscore');
lib/services/ngsi/ngsiService.js:const _ = require('underscore');
lib/services/ngsi/ngsiUtils.js:const _ = require('underscore');
lib/services/stats/statsRegistry.js:const _ = require('underscore');
lib/services/northBound/contextServer-NGSI-v2.js:const _ = require('underscore');
lib/services/northBound/deviceGroupAdministrationServer.js:const _ = require('underscore');
lib/services/northBound/restUtils.js:const _ = require('underscore');
[...]

PR #1119, opened 2021-10-19
Fix #858 - Remove request library

jason-fox:
- Replace `request` with `got`
- Add shim to mimic the old request callback interface
- Expose shim in API (for use downstream)
- Update documentation
- Update tests (replace `request` with a utility function)

PR #1170, opened 2022-01-03
Expose common libs

jason-fox:
This PR exposes 3 common libraries for use by IoT Agents

`async`, `express`  and `bodyParser` can be used in the same manner as the `request-shim`

Related https://github.com/telefonicaid/iotagent-node-lib/issues/1122


jason-fox:
I didn't include `underscore` in this list since it is likely to be removed https://github.com/telefonicaid/iotagent-node-lib/issues/1168 . By removing the duplicate dependencies you keep all agents in lock-step versioning - I could imagine a future improvement where further changes occur here once and once only (_for example an `express-shim.js` is introduced and all downstream IoT Agents  start using [Fastify](https://medium.com/@onufrienkos/express-vs-fastify-performance-4dd5d73e08e2))_

jason-fox:
Thanks to #1173 and friends use of `logops` is consistent, but it demonstrates the need to reduce the number of dependencies `underscore` or `lodash` is already a dependency-of-a-dependency since `logops`  uses it - you can check your `package-lock.json` for details. The question here is which of the dependencies (if any) should be exposed for use by downstream IoT Agents.

mapedraza:
This is an interesting thread. After discussing internally, we arrive to some general ideas, depending on what to do with library: 

1. **Exposing** libs in `iotagent-node-lib` for use by IoTAgents. As a general rule, it makes sense to expose in the IOTA-Lib to be used by te IOTA: 
- those libraries  that have configuration-based status (e.g. logops is a clear example) 
- those libraries where wrapping (encapsulating) is done in IOTA-Lib (e.g. request).

2. **Encapsulating** libs in `iotagent-node-lib` for use by IoTAgents. It makes sense for those libraries that have available multiples alternatives, there is a risk of deprecation, or common functions are used across the codebase in different files / modules (eg. request). 

Note that when encapsulating a library, it will be usually necessary to expose it as well (unless the encapsulation is done only for internal use in the `iotagent-node-lib` without needing to be used by the agents).

As an example for the previous dependencies: 

| Library | Exposed | Encapsulated |
|---------|---------|--------------|
| async   | No      | No           |
| logs    | Yes     | Yes          |
| express | ? *       | No           |
| bodyparser | No    | No      | 

\* Does `express` have configuration-based status?

This is not a general rule to be applied to all the dependencies, need to be review one by one (where is used the dependency, how difficult is to wrap the module and so on)

@jason-fox does it makes sense for you?

jason-fox:
Agreed. There need to be a clear list of exposed libraries. This needs explaining in the code comments **and* in the documentation. The code comments need to describe what goes above the line and what goes below the line and why e.g.:

```javascript
// These are common libraries which are exposed to be used by IoT Agents

exports.logModule = logger;  // Common configurable logger for all IoT Agents
exports.request = request; // IoT Agents can reuse standard logging shim
exports.async = async;  /// etc.
exports.express = require('express'); // etc.
exports.bodyParser = require('body-parser'); // etc.
```

If you can't justify something below the line it  shouldn't be exposed. The Telefónica Team need to make the call as to which deps get exposed.

jason-fox:
Even if something is not exposed, someone should run the following query - e.g. for `logops`  it would be:

https://github.com/search?q=org%3Atelefonicaid+%22logops%22%3A++extension%3Ajson&type=Code

and update the `xxxx` dependency across relevant repos to match the  `xxxx` dependency version used by the IoT Agent Node lib. You wouldn't get fully consistent alignment, but at least if you do it once then you'll start off with a common usage - it is then a matter of drift.

jason-fox:
I've just been running `npm audit` and `npm audit --production` across my IoT Agent forks, the node lib is `npm audit` currently:

```
found 16 vulnerabilities (11 moderate, 4 high, 1 critical) in 884 scanned packages
```

However this is almost all `devDependencies` - there are just two production issues, both in the `mongoose` package.

```
found 2 vulnerabilities (1 moderate, 1 critical) in 160 scanned packages
```

When running an IoT Agent, these two are cascaded down - for example [iotagent-isoxml](https://github.com/FIWARE/iotagent-isoxml):

```

┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Critical      │ Type confusion in mpath                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ mpath                                                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.8.4                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ iotagent-node-lib                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ iotagent-node-lib > mongoose > mpath                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://github.com/advisories/GHSA-p92x-r36w-9395            │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Moderate      │ Code Injection in mquery                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ mquery                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=3.2.3                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ iotagent-node-lib                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ iotagent-node-lib > mongoose > mquery                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://github.com/advisories/GHSA-45q2-34rf-mr94            │
└───────────────┴──────────────────────────────────────────────────────────────┘

found 2 vulnerabilities (1 moderate, 1 critical) in 160 scanned packages
```

jason-fox:
If I understand your criteria correctly, so far only the `logops` dependency is identified as to be encapsulated.

However, as a thought experiment, consider the following scenario regarding `express` -  at the moment, `express` is used for responding to the context broker. The current response  message is insecure as it is missing some common security headers which could be rectified by adding [helmet](https://www.npmjs.com/package/helmet), and could apply the following `setSecurityHeaders()` function using configurable settings:

```javascript
const helmet = require('helmet');
app.disable('x-powered-by');

app.use(
  helmet.dnsPrefetchControl({
    allow: process.env.IOTA_DNS_PREFETCH_ALLOW === 'true'
  })
);
app.use(helmet.expectCt());
app.use(
  helmet.frameguard({
    action: process.env.IOTA_FRAMEGUARD_ACTION || 'sameorigin'
  })
);
app.use(
  helmet.hsts({
    maxAge: process.env.IOTA_HTTPS_MAX_AGE || 15552000, 
    includeSubDomains: process.env.IOTA_INCLUDE_SUB_DOMAINS !== 'false', 
    preload: process.env.IOTA_PRELOAD === 'true'
  })
);
app.use(helmet.ieNoOpen());
app.use(helmet.noSniff());
app.use(
  helmet.permittedCrossDomainPolicies({
    permittedPolicies: process.env.IOTA_PERMITTED_POLICIES || 'none' 
  })
);
app.use(helmet.referrerPolicy());
app.use(helmet.xssFilter());
```

Once something like this is done, the `setSecurityHeaders()` function could legitimately be exposed so downstream IoT Agents using `express` (like Ultralight or JSON could benefit) from dependency`helmet` updates indirectly.

jason-fox:
Also related to https://github.com/telefonicaid/iotagent-node-lib/pull/1170#issuecomment-1011899738 `.github/workflows/ci.yml` should be amended to output the production vulnerabilities listing just in case.

```diff
   steps:
      - name: Git checkout
        uses: actions/checkout@v2
      - name: 'Install Node.js ${{ matrix.node-version }}'
        uses: actions/setup-node@v1
        with:
          node-version: '${{ matrix.node-version }}'
      - name: 'Unit Tests with Node.js ${{ matrix.node-version }}'
        run: |
          npm install
+         npm audit --production
          npm test
```

99% of the time this will just make the logs slightly longer, but it would help diagnose issues like #1173

mapedraza:
> If I understand your criteria correctly, so far only the `logops` dependency is identified as to be encapsulated.
> 
> However, as a thought experiment, consider the following scenario regarding `express` - at the moment, `express` is used for responding to the context broker. The current response message is insecure as it is missing some common security headers which could be rectified by adding [helmet](https://www.npmjs.com/package/helmet), and could apply the following `setSecurityHeaders()` function using configurable settings:
> 
> ```js
> const helmet = require('helmet');
> app.disable('x-powered-by');
> 
> app.use(
>   helmet.dnsPrefetchControl({
>     allow: process.env.IOTA_DNS_PREFETCH_ALLOW === 'true'
>   })
> );
> app.use(helmet.expectCt());
> app.use(
>   helmet.frameguard({
>     action: process.env.IOTA_FRAMEGUARD_ACTION || 'sameorigin'
>   })
> );
> app.use(
>   helmet.hsts({
>     maxAge: process.env.IOTA_HTTPS_MAX_AGE || 15552000, 
>     includeSubDomains: process.env.IOTA_INCLUDE_SUB_DOMAINS !== 'false', 
>     preload: process.env.IOTA_PRELOAD === 'true'
>   })
> );
> app.use(helmet.ieNoOpen());
> app.use(helmet.noSniff());
> app.use(
>   helmet.permittedCrossDomainPolicies({
>     permittedPolicies: process.env.IOTA_PERMITTED_POLICIES || 'none' 
>   })
> );
> app.use(helmet.referrerPolicy());
> app.use(helmet.xssFilter());
> ```
> 
> Once something like this is done, the `setSecurityHeaders()` function could legitimately be exposed so downstream IoT Agents using `express` (like Ultralight or JSON ) could benefit from dependency`helmet` updates indirectly.

It is right, `logops` is a good candidate to be encapsulated. 

In the case of express, it could be a good candidate to be exposed once it is initialized. Since the typical usage an IoTA is to be deployed in the same network of the rest of the componentes (i.e. Context Broker), it is not planed to include helmet dependency in short term. Nevertheless is a good use case example to take into account.

fgalan:
> The encapsulation should also take into account the previous usage in order to not alter the usage of the codebase. As an example, #1119 the wrapper was not having the exact behaviour of the previous library as raised described in #1178. Fortunately, it was detected in E2E tests. In future encapsulations (and specially replacing dependencies) we should ensure that the module works exactly as before.

This is very important. Unfortunately we missed this with the request-shim work at PR #1119 and reverting it back would involve significative work, but we should learn for the future and avoid changes in the API usage of the wrapped module.

jason-fox:
I think the way forward is clear now. I'll close this and the related Proof of Concept PR https://github.com/telefonicaid/iotagent-ul/pull/526 unmerged and raise simpler PRs one-by-one as necessary. It looks like the agreed way to do this is one library at a time.

`iotAgentLib.logModule` already is exposed, so the next logical step would be to remove the unnecessary `logOps` dependency across the board from IoT Agents.

fgalan:
Maybe the conclusions of this PR should be moved to parent issue https://github.com/telefonicaid/iotagent-node-lib/issues/1122

PR #1181, opened 2022-01-18
[PoC] Check DELETE expectations for mocks 1178

mapedraza:
This PR is created with the aim of modify the expectations to (this should break the tests because in some cases the DELETE is sending some data as described in #1178)

### DIFF
--- a/lib/services/devices/registrationUtils.js
+++ b/lib/services/devices/registrationUtils.js
@@ -176,7 +176,6 @@ function sendUnregistrationsNgsiLD(deviceData, callback) {
     const options = {
         url: cbHost + '/ngsi-ld/v1/csourceRegistrations/' + deviceData.registrationId,
         method: 'DELETE',
-        json: true,
         headers: {
             'fiware-service': deviceData.service,
             'fiware-servicepath': deviceData.subservice,
--- a/lib/services/devices/registrationUtils.js
+++ b/lib/services/devices/registrationUtils.js
@@ -136,7 +136,6 @@ function sendUnregistrationsNgsi2(deviceData, callback) {
     const options = {
         url: cbHost + '/v2/registrations/' + deviceData.registrationId,
         method: 'DELETE',
-        json: true,
         headers: {
             'fiware-service': deviceData.service,
             'fiware-servicepath': deviceData.subservice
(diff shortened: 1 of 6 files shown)


mapedraza:
Now the test edited ins failing as shown in [tests results log](https://github.com/telefonicaid/iotagent-node-lib/runs/4855922694?check_suite_focus=true#step:5:1626) because it is receiving `"body": "true"`

>  message: 'Nock: No match for request {\n  "method": "DELETE",\n  "url": "http://192.168.1.1:1026/v2/registrations/6319a7f5254b05844116584d",\n  "headers": {\n    "user-agent": "got (https://github.com/sindresorhus/got)",\n    "fiware-service": "smartgondor",\n    "fiware-servicepath": "gardens",\n    "accept": "application/json",\n    "content-type": "application/json",\n    "content-length": "4",\n    "accept-encoding": "gzip, deflate, br"\n  },\n  **"body": "true"\n**}'

mapedraza:
In order to check if expectations would work if the request is not sending data, next step is to remove the body, fixing `json : true` in the code as needed in #1178

mapedraza:
patch generated from #1178 applied to this branch. Now, the edited test should not fail

lib/services/devices/registrationUtils.js
```diff
@@ -136,7 +136,6 @@ function sendUnregistrationsNgsi2(deviceData, callback) {
     const options = {
         url: cbHost + '/v2/registrations/' + deviceData.registrationId,
         method: 'DELETE',
-        json: true,
```
AlvaroVega, line 139: This was already deleted by https://github.com/telefonicaid/iotagent-node-lib/pull/1178

mapedraza: This PR is a PoC. Changes in #1178 were included here in order to check if changes on tests were capturing the problem exposed and, after applying the patch, if it continues or not.

test/unit/ngsiv2/provisioning/device-registration_test.js
```diff
@@ -290,7 +290,7 @@ describe('NGSI-v2 - IoT Agent Device Registration', function () {
             contextBrokerMock.post('/v2/entities?options=upsert').reply(204);
 
             contextBrokerMock
-                .delete('/v2/registrations/6319a7f5254b05844116584d')
+                .delete('/v2/registrations/6319a7f5254b05844116584d',"")
```
mapedraza, line 293: The important change in this PR is this change, which checks the body of the DELETE request (in this particular case, checks if it is empty)

mapedraza:
After creating https://github.com/telefonicaid/iotagent-node-lib/pull/1183, this PR can be closed since the changes needed to enhance the expectations were made in https://github.com/telefonicaid/iotagent-node-lib/pull/1183

Issue #1189, opened 2022-02-01
context log used in plugins is not filling serv/subserv properly


AlvaroVega:
Analize/Fix

fgalan:
Fixed in PR https://github.com/telefonicaid/iotagent-node-lib/pull/1192

PR #1192, opened 2022-02-02
fill proper service and subservice

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/1189


lib/plugins/expressionParser.js
```diff
@@ -28,8 +28,9 @@
 const Parser = require('jison').Parser;
```
fgalan, line 28: CNR entry should be added (do we have a related issue?). Seems to be an important fix :)

What about the IOTA code? If we have copied the same pattern, the bug would also exists there...

AlvaroVega: done in https://github.com/telefonicaid/iotagent-node-lib/pull/1192/commits/f65e2b4e79e09c772c31e31620a40618172fa6ef

fgalan (approved):
LGTM

Issue #1211, opened 2022-03-23
Bad command type provisioned when entity_id is duplicated


AlvaroVega:
Provisioning:
2 groups with several commands
2 devices with different device_id, entity_type but same entity_id

command of second device is of type "string" instead of "command"

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1212

PR #1212, opened 2022-03-23
fix: search device and group for a command using entity type when provided

AlvaroVega:
fix for https://github.com/telefonicaid/iotagent-node-lib/issues/1211


fgalan (approved):
LGTM

Issue #1216, opened 2022-03-29
error log about search by group with fields "type":"/iot/json"


AlvaroVega:
time=2022-03-29T08:15:56.235Z | lvl=DEBUG | corr=d85d99f0-04e0-46bb-b561-c6a8e4e2d455 | trans=d85d99f0-04e0-46bb-b561-c6a8e4e2d455 | op=IoTAgentNGSI.MongoDBGroupRegister | from=n/a | srv=n/a | subsrv=n/a | msg=Looking for group params ["type"] with queryObj {"type":"/iot/json"} | comp=IoTAgent
time=2022-03-29T08:15:56.240Z | lvl=DEBUG | corr=d85d99f0-04e0-46bb-b561-c6a8e4e2d455 | trans=d85d99f0-04e0-46bb-b561-c6a8e4e2d455 | op=IoTAgentNGSI.MongoDBGroupRegister | from=n/a | srv=n/a | subsrv=n/a | msg=Device group for fields [["type"]] not found: [{"type":"/iot/json"}] | comp=IoTAgent

time=2022-03-29T07:52:10.450Z | lvl=ERROR | corr=d85d99f0-04e0-46bb-b561-c6a8e4e2d455 | trans=d85d99f0-04e0-46bb-b561-c6a8e4e2d455 | op=IoTAgentNGSI.Alarms | from=n/a | srv=smartcity | subsrv=/ | msg=Raising [MONGO-ALARM]: {"name":"DEVICE_GROUP_NOT_FOUND","message":"Couldn\t find device group for fields: [\"type\"] and values: {\"type\":\"/iot/json\"}","code":404} | comp=IoTAgent

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/blob/6dbfaaf719323e90989459769f34b81c32260b2f/lib/services/northBound/contextServerUtils.js#L94

fgalan:
Fixed by PR #1217

PR #1217, opened 2022-03-29
fix bad query searching for group using resource instead of type

AlvaroVega:
fix for https://github.com/telefonicaid/iotagent-node-lib/issues/1216


fgalan (approved):
LGTM

PR #1209, opened 2022-03-17
add expression, payloadType and contentType to command model

AlvaroVega:

PR #1221, opened 2022-03-31
FIX api.md

fgalan:
Realized of this while reviewing PR https://github.com/telefonicaid/iotagent-node-lib/pull/1209

As far as I understand the attribute configurations applies to groups and devices.


AlvaroVega (approved):
LGTM

PR #1225, opened 2022-04-04
Exclude processing all attrs expression when current attr to update is of type commandStatus of commandResult

AlvaroVega:
Due to some expressions of attributes could not be executed in a non measure context (i.e. command response from a device), resulting in errors or filling nulls in attrs

- [x] And the same for commandResult type

This PRs enables commands of devices with another attributes using expressions could transit to PENDING and DELIVERED status

### DIFF
--- a/lib/plugins/expressionPlugin.js
+++ b/lib/plugins/expressionPlugin.js
@@ -31,6 +31,7 @@ const config = require('../commonConfig');
 /* eslint-disable no-unused-vars */
 const logger = require('logops');
 const errors = require('../errors');
+const constants = require('../constants');
 const context = {
     op: 'IoTAgentNGSI.expressionPlugin'
 };
@@ -134,7 +135,7 @@ function update(entity, typeInformation, callback) {
         let attsArray = utils.extractAttributesArrayFromNgsi2Entity(entity);
         // Exclude processing all attr expressions when current attr is of type 'commandStatus' or 'commandResult'
         const attsArrayFiltered = attsArray.filter((obj) => {
-            return !['commandStatus', 'commandResult'].includes(obj.type);
+            return ![constants.COMMAND_STATUS, constants.COMMAND_RESULT].includes(obj.type);
         });
         if (attsArrayFiltered.length > 0) {
             attsArray = processEntityUpdateNgsi2(attsArrayFiltered);
--- a/lib/plugins/expressionPlugin.js
+++ b/lib/plugins/expressionPlugin.js
@@ -132,9 +132,9 @@ function update(entity, typeInformation, callback) {
     try {
         logger.debug(context, 'expressionPlugin entity %j', entity);
         let attsArray = utils.extractAttributesArrayFromNgsi2Entity(entity);
-        // Exclude processing all attrs expression when current attr is of type 'commandStatus'
+        // Exclude processing all attr expressions when current attr is of type 'commandStatus' or 'commandInfo'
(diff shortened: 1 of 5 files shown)


lib/plugins/expressionPlugin.js
```diff
@@ -131,7 +132,13 @@ function update(entity, typeInformation, callback) {
     try {
         logger.debug(context, 'expressionPlugin entity %j', entity);
         let attsArray = utils.extractAttributesArrayFromNgsi2Entity(entity);
-        attsArray = processEntityUpdateNgsi2(attsArray);
+        // Exclude processing all attr expressions when current attr is of type 'commandStatus' or 'commandInfo'
```
fgalan, line 135:
```suggestion
        // Exclude processing all attr expressions when current attr is of type 'commandStatus' or 'commandResult'
```

fgalan:
We have been talking (@AlvaroVega @fgalan ) by skype about this PR. Some conclusions:

* In https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/usermanual.md#iotagentlibsetcommandresult documentation, maybe the usage of `commandStatus` and `commandResult` types should be mentioned, when describing the `_status` and `_info` side attributes.
* In the place of library doc that describes registrations (if any) the usage of `command` type for registering commands
* It seems there is a typo in documentation, mentioning `_result` instead of `_status` in some places. Maybe in the old past we used `_result` but documentation was not fully fixed.
* IOTA-JSON and IOTA-UL documentation should be reviewed regarding this

AlvaroVega:
There are some info at: https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/northboundinteractions.md#scenario-3-commands:


Another attribute will be used as the result attribute. This attribute will be updated from the IoTAgent, and its value stored in the Context Broker. This attribute will contain the result of the command (this result can be information in case the command was a "information retrieval" command or the result of an action if it was an "actuator command"). Typically, the name of this attribute will be the same of the input attribute, with an additional sufix (_info).

AlvaroVega:
According with: https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/northboundinteractions.md#result-reporting

Result reporting

Once the IoT Agent has executed the command or retrieved the information from the device, it reports the results to the Context Broker, with an updateContext (P1):

curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" -H "Fiware-Service: workshop" \
  -H "Fiware-ServicePath:  /iota2ngsi " -H "X-Auth-Token: <token>" -d '{
    "entities": [
        {
            "type": "device",
            "isPattern": "false",
            "id": "Dev0001",
            "switch_info": {
                "type": "command_info",
                "value": "Switched successfully!"
            },
            "switch_status": {
                "type": "command_status",
                "value": "OK"
            }
        }
    ],
    "actionType": "update"
} ' "https://<platform-ip>:10027/v2/op/update"

This update does not modify the original command attribute, but two auxiliary attributes, that are not provided by the IoT Agent (usually, those attributes has the same name as the command, with an added suffix).

The Context Broker replies to the IoT Agent with a R1 payload (200 OK):

[
    {
        "type": "device",
        "id": "Dev0001",
        "switch_info": {
            "type": "command_info",
            "value": ""
        },
        "switch_status":  {
            "type": "command_status",
            "value": ""
        }
    }
]

fgalan (approved):
LGTM

Issue #1235, opened 2022-04-21
explicitAttrs when is []


AlvaroVega:
given a device with:
```
explicitAttrs : object?['status']:[]
```
and an attribute like:

Protocol parameter | Device attribute | Type | Expression | Entity Type | Entity Name
-- | -- | -- | -- | -- | --
fake1 | status | Text | object.a?object.a:null

sending a measure:
```
curl -i -X POST 'http://localhost:7897/iot/json?i=disp1&k=6idn7z9cba3o57tgoascazdwd' -d '{ "fake1": "OK", "object3": { "a": "3"}}' -H 'content-type: application/json'
```
Should not update with null status

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1236

PR #1180, opened 2022-01-17
extend ExplicitAttrs to use JSON/JEXL

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/1179

- [x] Test added
- [x] Doc describe new ExplicitAttrs flavor

PR #1236, opened 2022-04-21
Fix/explicit attrs when is an empty array

AlvaroVega:
In that case (explictAttrs = []) none attribute should be sent to CB.

issue https://github.com/telefonicaid/iotagent-node-lib/issues/1235
related with https://github.com/telefonicaid/iotagent-node-lib/pull/1180


fgalan (approved):
LGTM

PR #1237, opened 2022-04-22
upgrade async dep from 2.6.2 version to 2.6.4 due to security vulnerability (CWE-1321)

AlvaroVega:
A vulnerability exists in Async through 3.2.1 for 3.x and through 2.6.3 for 2.x (fixed in 3.2.2 and 2.6.4), which could let a malicious user obtain privileges via the mapValues() method.


fgalan (approved):
LGTM

PR #1247, opened 2022-05-12
 identify each flow of mongo alarms about device group

AlvaroVega:


fgalan:
Documentation about mongo alarms should be modified to explain the suffix thing.

fgalan:
Documentation about mongo alarms should be modified to explain the suffix thing.

PR #1249, opened 2022-05-16
fix: Lower case headers for NGSILD Tenant and NGSILD Path

wiikip:
req.headers only contains lowercased headers.
Calling req.headers["NGSILD-Tenant"] will always return undefined. The lib cannot have the expected comportement.
Replacing key with "ngsild-tenant" is fixing the problem. I tested it with iot-agent-json lib ( custom image wiikip/iot-agent-json )

### DIFF
--- a/lib/services/northBound/contextServerUtils.js
+++ b/lib/services/northBound/contextServerUtils.js
@@ -41,8 +41,8 @@ const commands = require('../commands/commandService');
  * @return {String}                 The Tenant decribed in the request headers
  */
 function getLDTenant(req) {
-    if (req.headers['NGSILD-Tenant']) {
-        return req.headers['NGSILD-Tenant'];
+    if (req.headers['ngsild-tenant']) {
+        return req.headers['ngsild-tenant'];
     } else if (req.headers['fiware-service']) {
         return req.headers['fiware-service'];
     }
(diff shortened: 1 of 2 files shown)


lib/services/northBound/contextServerUtils.js
```diff
@@ -41,8 +41,8 @@ const commands = require('../commands/commandService');
  * @return {String}                 The Tenant decribed in the request headers
```
fgalan, line 41: An entry in the CHANGES_NEXT_RELEASE file corresponding to the fix done in this PR should be included.

wiikip: Just updated it

fgalan: Fixed in [842af08](https://github.com/telefonicaid/iotagent-node-lib/pull/1249/commits/842af087eb7ca4ba342b62958345643c9a0cf52e)

fgalan:
@jason-fox what do you think on this PR? LGTM from  your side?

jason-fox:
Yes that is a bug, yes that is a fix, but according to [this](https://stackoverflow.com/questions/60855411/req-header-vs-req-headers-in-express) a better fix would be to use `req.header('Content-Type');` rather than `req.headers['Content-Type'];`

That would allow **the code** to continue using mixed case letters which are easier to read and also align with the definitions found in the ETSI NGSI-LD specification:

e.g something like this:

```javascript
function getLDTenant(req) {
    if (req.header('NGSILD-Tenant')) {
        return req.header('NGSILD-Tenant');
    } else if (req.header('fiware-service')) {
        return req.header('fiware-service');
    }
    return config.getConfig().contextBroker.fallbackTenant;
}


function getLDPath(req) {
    if (req.header('NGSILD-Path')) {
        return req.header('NGSILD-Path');
    } else if (req.header('fiware-servicepath')) {
        return req.header('fiware-servicepath');
    }
    return config.getConfig().contextBroker.fallbackPath;
}
```

wiikip:
Did not know about this function but it s a way more elegant fix. Just updated

jason-fox:
LGTM

mapedraza (approved):
LGTM

PR #1257, opened 2022-07-12
set jexl as default parser

AlvaroVega:
At command exectuions typeInformation is just device (without group information)


fgalan (approved):
LGTM

PR #1260, opened 2022-07-12
[WIP] add id type service and subservice to context expression of multientity plugin

AlvaroVega:


doc/expressionLanguage.md
```diff
@@ -342,6 +342,21 @@ Will now generate the following NGSI v2 payload:
 }
 ```
 
+## Available keys for all Expressions
+
+Not just for measure transfomration (including multientity) are expressions availables, also for:
```
fgalan, line 347:
```suggestion
Apart from measurement transformations (including multientity cases), expressions can be used in the following cases:
```

PR #1262, opened 2022-07-12
Add NGSI-LD LanguageProperty support

jason-fox:
`LanguageProperty` has been defined in the NGSI-LD spec since 1.4.1. This PR correctly maps LanguageProperty to a `languageMap` attribute which is defined in the core `@context` as a JSON-LD `@container: @lang`. 

`LanguageProperty` is most likely to be a static attribute, but active measures can also be supported - it is basically the same idea as mapping `Relationship` and the `object` attribute.

- Update Docs
- Update Unit Test
- Add `LanguageProperty` exception clause to `switch`


fgalan:
Lint javascript check failed.

mapedraza (approved):
LGTM

PR #1276, opened 2022-08-04
Ensure GET `http://{{iot-agent}}/ngsi-ld/v1/entities/<entity-id>` returns a valid response

jason-fox:
In NGSI-LD, normally `http://{{iot-agent}}/ngsi-ld/v1/entities/<entity-id>` will return lazy attributes , however, I just noticed a bug related to commands. If a device had registered commands, but no registered lazy attribute it was failing with a null pointer and the response was not valid:

- Defensive Programming - add null check to attributes
- Add Test.

`http://{{iot-agent}}/ngsi-ld/v1/entities/<entity-id>` is one of the **valid** NGSI-LD endpoints defined in the 1.5.3 spec which the IoT Agent should service. 

Admittedly, ideally the context broker **shouldn't** need to call the IoT Agent at all in this case (with commands but no lazy attributes) but since  is a valid request when a device has lazy attributes or when a device has lazy attributes **and** commands, it shouldn't error in this case, but give a valid empty response.

### DIFF
--- a/lib/services/northBound/contextServer-NGSI-LD.js
+++ b/lib/services/northBound/contextServer-NGSI-LD.js
@@ -290,7 +290,7 @@ function handleQueryNgsiLD(req, res, next) {
 
         if (device.staticAttributes) {
             let selectedAttributes = [];
-            if (attributes === undefined || attributes.length === 0) {
+            if (attributes === null || attributes === undefined || attributes.length === 0) {
                 selectedAttributes = device.staticAttributes;
             } else {
                 selectedAttributes = device.staticAttributes.filter(inAttributes);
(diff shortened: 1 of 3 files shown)


lib/services/northBound/contextServer-NGSI-LD.js
```diff
@@ -290,7 +290,7 @@ function handleQueryNgsiLD(req, res, next) {
```
fgalan, line 290: The fix in this PR should be described in CHANGES_NEXT_RELEASE file

jason-fox: Fixed f8380a7

fgalan (approved):
LGTM

PR #1278, opened 2022-08-10
[WIP] increase bodyparser limit from 100kbs to 1mb

AlvaroVega:
With a huge number of groups in iotagent-json, iotagent-manager is getting:

time=2022-08-10T08:00:42.937Z | lvl=DEBUG | corr=6a8d75e6-e10f-45ea-8ab0-730ca1334f69 | trans=6a8d75e6-e10f-45ea-8ab0-730ca1334f69 | op=IoTAManager.Redirector | from=n/a | srv=n/a | subsrv=n/a | msg=Error [PayloadTooLargeError] handling request: request entity too large | comp=IoTAgentManager

So we need try, as woraround, increase limits:

https://stackoverflow.com/questions/50988891/payloadtoolargeerror-with-limit-set

https://github.com/expressjs/body-parser

limit

Controls the maximum request body size. If this is a number, then the value specifies the number of [bytes](https://www.npmjs.com/package/bytes); if it is a string, the value is passed to the bytes library for parsing. Defaults to '100kb'.

### DIFF
--- a/lib/services/northBound/northboundServer.js
+++ b/lib/services/northBound/northboundServer.js
@@ -56,8 +56,8 @@ function start(config, callback) {
     northboundServer.app.set('port', config.server.port);
     northboundServer.app.set('host', config.server.host || '0.0.0.0');
     northboundServer.app.use(domainUtils.requestDomain);
-    northboundServer.app.use(bodyParser.json({ limit: '200mb' }));
-    northboundServer.app.use(bodyParser.json({ limit: '200mb', type: 'application/*+json' }));
+    northboundServer.app.use(bodyParser.json({ limit: '1mb' }));
+    northboundServer.app.use(bodyParser.json({ limit: '1mb', type: 'application/*+json' }));
 
     if (config.logLevel && config.logLevel === 'DEBUG') {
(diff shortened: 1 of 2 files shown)


lib/services/northBound/northboundServer.js
```diff
@@ -56,8 +56,8 @@ function start(config, callback) {
     northboundServer.app.set('port', config.server.port);
     northboundServer.app.set('host', config.server.host || '0.0.0.0');
     northboundServer.app.use(domainUtils.requestDomain);
-    northboundServer.app.use(bodyParser.json());
-    northboundServer.app.use(bodyParser.json({ type: 'application/*+json' }));
+    northboundServer.app.use(bodyParser.json({ limit: '1mb' }));
+    northboundServer.app.use(bodyParser.json({ limit: '1mb', type: 'application/*+json' }));
```
fgalan, line 60: Maybe this should be unhardwired to a configuration parameter?

AlvaroVega: Is there any other similiar configuration parameter to do it in a similar way?

AlvaroVega:
https://github.com/telefonicaid/iotagent-manager/pull/296

Issue #1292, opened 2022-09-13
Wrong management of default entity_name at multientity with JEXL


mrutid:
When a default entity name ("type:device_id") or any "wrong JEXL expression" literal is used as entity name, an exception is raised. All the following will fail when a measure is received:

 "attributes" : [ 
        {
            "name" : "a",
            "type" : "Number",
            "entity_name" : "type:deviceid",
            "entity_type" : "myType"
        },
        {
            "name" : "b",
            "type" : "Number",
             "entity_type" : "myType"
        }
    ]

Wrong JEXL expression should behave same as <null> result

fgalan:
Fixed in PR https://github.com/telefonicaid/iotagent-node-lib/pull/1293

PR #1293, opened 2022-09-13
Avoid Exception on expression check

mrutid:
1 LD test is not passing

Issue #1292


CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
 - Set Nodejs 12 as minimum version in packages.json (effectively removing Nodev10 from supported versions)
+- New INVALID EXPRESSION policy at JEXL expression in attributes and entity_names. Defaul values will be propagated and a warn will be logged instead of exception ([#1292](https://github.com/telefonicaid/iotagent-node-lib/issues/1292))
```
fgalan, line 2:
```suggestion
- Fix: INVALID EXPRESSION policy at JEXL expression in attributes and entity_names (default values will be propagated and a warn will be logged instead of exception) (#1292)
```

fgalan (approved):
LGTM

PR #1277, opened 2022-08-04
Add NGSI-LD Unsupported Endpoints

jason-fox:
-  GET `/ngsi-ld/v1/entities`
-  POST `/ngsi-ld/v1/entities`
-  DELETE `/ngsi-ld/v1/entities/<entity-id>`
-  DELETE `ngsi-ld/v1/entities/<entity-id>/attrs/<attr-name>`

Are valid NGSI-LD endpoints, but IoT Agents will never support them. Ensure that they return valid NGSI-LD error responses instead. This protects the IoT Agent and also informs the Context Broker of the reason why. For example, at the moment Scorpio assumes full federated `/ngsi-ld/v1/entities` is supported when it isn't. see: telefonicaid/iotagent-ul/issues/573

Issue #1166, opened 2021-12-22
Refactor entities-NGSI-v2.js module


mrutid:
A profound refactor of the module is needed, starting with **sendUpdateValueNgsi2** method. 

Currently there are too much branching due to plugin application (mainly multi-entity) which could be addressed in an uniform way. For example everything could be considered "multientity" and use just the bulk method of the NGSI API.

The plugin architecture should be revised also: Name Mapping, Expression and Multi-entity plugin are not optional or pluggable. If we consider their business logic from the beginning in an integrated way many of the current complexity will disappear (waterfalling synchronous methods and decorating shared objects to later on remove decorations seems weird). 

A better world is possible. 

This may apply also to the brother module: entities-NGSI-LD.js

AlvaroVega:
Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1314

PR #1302, opened 2022-11-22
Update to offer NGSI-LD 1.6.1. Registration

jason-fox:
Related to https://github.com/telefonicaid/iotagent-ul/issues/573#issuecomment-1203172196

The [NGSI-LD 1.6.1 Specification](https://www.etsi.org/deliver/etsi_gs/CIM/001_099/009/01.06.01_60/gs_CIM009v010601p.pdf), clarifies the definition of registrations and defines `operations` and `mode` attributes to be supplied within NGSI-LD **CSourceRegistrations**.  A default Registration is considered to be `federationOps/inclusive` (which is what [Scorpio 3](https://github.com/ScorpioBroker/ScorpioBroker/releases) already offers and [Orion-LD 1.2](https://github.com/FIWARE/context.Orion-LD/releases) is aiming for.

This is a potentially **breaking change** for provisioning NGSI-LD devices - the existing functionality  aligns with the Orion-LD 1.1 prototype only - there is little point in continuing to offer the provisional NGSI-LD 1.3.1. interface.  Additionally `properties` was updated to `propertyNames` in NGSI-LD 1.4.1, Orion-LD offers a backwards compatible option, other context brokers may not.

#### How are devices affected?

1) Basic Sensors are not affected by this change
2) Sensors with lazy attributes should  register as `retrieveOps/exclusive`
3) Actuation command attributes should  register as `updateOps/exclusive`

The default `federationOps/inclusive`  offers a superset of these commands so previously provisioned devices will cause unnecessary polling traffic to the IoT Agent (which the IoT Agent already rejects as unsupported see #1277). 

This really isn't a big issue for this library (and the fix is really small) but for semantic versioning it would make sense to update the semantic versioning from 2.x.x to 3.x.x when it lands to indicate the move from the "experimental" operations defined in #849  et seq. which work with Orion-LD only, up to the well defined common interface supported by all NGSI-LD brokers.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,2 @@
-
+- Update to offer NGSI-LD 1.6.1. Registrations #1302
+- Document removal of support for NGSI-LD 1.3.1 Interface
--- a/doc/deprecated.md
+++ b/doc/deprecated.md
@@ -18,7 +18,7 @@ A list of deprecated features and the version in which they were deprecated foll
 -   Support to Node.js v8 in iotagent-node-lib 2.12.0 (finally removed in 2.13.0)
 -   Support to Node.js v10 in iotagent-node-lib 2.15.0 (finally removed in 2.16.0)
 -   Support to Node.js v12 in iotagent-node-lib 2.24.0 (finally removed in 2.25.0)
--   Support to NGSI-LD v1.3 in iotagent-node-lib 2.25.0
+-   Support to NGSI-LD v1.3 in iotagent-node-lib 2.25.0 (finally removed in 2.26.0)
 
 The use of Node.js v14 is highly recommended.
 
@@ -46,3 +46,4 @@ The following table provides information about the last iotagent-node-lib versio
 | Support to Node.js v8  | 2.12.0                                                | April 7th, 2020               |
 | Support to Node.js v10 | 2.15.0                                                | February 18th, 2021           |
 | Support to Node.js v12 | 2.24.0                                                | September 2nd, 2022           |
+| Support to NGSI-LD 1.3 | 2.25.0                                                | January 24th, 2023          |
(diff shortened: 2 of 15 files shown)


jason-fox:
An explanation of the differences between the various modes of operation can be found [here](https://www.youtube.com/watch?v=ebk14k5dNxo)

jason-fox:
This PR will remain in **DRAFT** until tested against compliant 1.6.1 versions of Orion-LD and Scorpio. Looking at the comment in the Ultralight Issue, this is the current state:

1. The ETSI NGSI-LD committee defines what should happen across brokers. ✅ 
2. The Scorpio codebase is updated to accept registrants which are not full brokers **IN PROGRESS**
3. The Orion-LD codebase is updated to accept true federations **IN PROGRESS**
4. The IoT Agent Node lib is updated to provide whatever flags are necessary when making a registration to describe its implementation as a partial broker. **THIS PR** - completed in DRAFT
...
6. A stable release of the IoT Agent Node lib is completed and added to the IoT Agent for JSON
7. A stable release of the IoT Agent for JSON is completed.

jason-fox:
@fgalan @mapedraza - as you can see the change here is pretty minor, but I'd like to kick off a conversation about how you propose to **deprecate** the existing provisional 1.3.1 NGSI-LD registration. Would you like an entry added to the [deprecated features list](https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/deprecated.md) - what sort of timelines would you envisage etc. etc.

mapedraza:
With respect to how to deprecate a feature, our modus operandi is to add that feature to the [deprecated features list](https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/deprecated.md) and keep it in the component for at least one release, announcing it in the changelog of the new version release, and giving at least a one-release window before removing it completely from the code base. That said, from Telefónica's side there is no problem with removing the feature completely without leaving the grace period.

As for increasing the version number, we are waiting to increase the major version number to 3.X.X when refactoring the plugins architecture and entities files (https://github.com/telefonicaid/iotagent-node-lib/issues/1166)

mapedraza:
CHANGES_NEXT_RELEASE entry is missing

PR #1303, opened 2022-11-30
Add deprecation notice

jason-fox:
Related #1302 - You should provide a heads up that the supported NGSI-LD interface will no longer accept the 1.3.1 `property` attribute in the `CSourceRegistration` at some point. This is just preparatory work before that PR can land at some time in the future.


jason-fox:
If you find this text acceptable, I can write something similar across all the other individual IoT Agent docs.

fgalan:
> If you find this text acceptable, I can write something similar across all the other individual IoT Agent docs.

I think is ok. Please go ahead and when we have all the PRs we will merge all them at the same moment.

jason-fox:
@fgalan - 4 more doc changes added.

PR #1321, opened 2023-03-21
Task/clean after plugins refactor

AlvaroVega:
This branch starts with iota-node-lib plugins refactor https://github.com/telefonicaid/iotagent-node-lib/pull/1314 and cleans all old code form plugins which is not necessary after that refactor.

PR #1314, opened 2023-02-24
Task/plugins refactor: alias, expressions, multientity and timestamp plugins

AlvaroVega:
Issue https://github.com/telefonicaid/iotagent-node-lib/issues/1195

This PR is about alias, expressions, multientity, timestamp  plugins. This new implementation allow remove alias, expressions, multientity, timestamp  plugins middelwares.
This PR still does not remove old stuff about plugins due ngsi-ld implementation and tests.

Test that were modified:

JSON reordered (but equivalent):
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityPlugin5.json
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityPlugin8.json
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityTimestampPlugin1.json
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityTimestampPlugin2.json

POST [batch 1 entity] -> PATCH [1 entity]
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityTimestampPlugin4.json

REMOVE ATT METADATA timestamp: timestamp is provided ad measure and other tests does not include it
test/unit/ngsiv2/examples/contextRequests/updateContextProcessTimestamp.json

REMOVE empty new multi entities (according with test not all entities are expected!)
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityPlugin6.json
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityPlugin7.json

ADD empty base entity
test/unit/ngsiv2/examples/contextRequests/updateContextMultientityPlugin11.json

The following plugin tests where modified:
- event: https://github.com/telefonicaid/iotagent-node-lib/blob/master/test/unit/ngsiv2/plugins/event-plugin_test.js
- translation: https://github.com/telefonicaid/iotagent-node-lib/blob/master/test/unit/ngsiv2/plugins/translation-inPlugins_test.js

This PR also fixes bug  https://github.com/telefonicaid/iotagent-json/issues/697 and https://github.com/telefonicaid/iotagent-node-lib/issues/1291

Pending tasks (for further PRs?)
- [ ] Removed unnecessary stuff from plugins (depends on ngsi-ld updates)
- [ ] Updated doc (depends on previous task)

PR #1322, opened 2023-03-21
Refactor NGSI-LD processing to align with alias, multientity, expressions and timestamp plugin changes

jason-fox:


jason-fox:
The drop in coverage nicely aligns with the files that @AlvaroVega  is editing in #1321. Of course since #1314 only factored out the NGSI-v2 side, it didn't experience the coverage drop itself.

PR #1339, opened 2023-04-18
Set JEXL as default expression language + Deprecate Legacy expressions

mapedraza:
This PR replace the default behaviour, when the language expression is not selected or defined, to JEXL instead of the LEGACY mode. 

*Edit:* issue https://github.com/telefonicaid/iotagent-node-lib/issues/1340


mapedraza:
These 2 parts of the code need to be amended:

https://github.com/telefonicaid/iotagent-node-lib/blob/7b7c9506bec8db6a29ac8ef3a193e969f09334e6/lib/fiware-iotagent-lib.js#L128-L136

https://github.com/telefonicaid/iotagent-node-lib/blob/7b7c9506bec8db6a29ac8ef3a193e969f09334e6/lib/plugins/expressionPlugin.js#L77-L94

mapedraza:
> And probably, some expression tests using legacy parser will fail since the default method could be not specified

After commit https://github.com/telefonicaid/iotagent-node-lib/pull/1339/commits/448a324c9e78ab9867754799aac61e7f91fa6449 we will see if they fail

mapedraza:
At this point (https://github.com/telefonicaid/iotagent-node-lib/pull/1339/commits/a15791af60f44715803b5727cb151daab3068f8a) everything is working apart from bidirectionality

After some research, it seems it is only compatible with LEGACY expressions. The part of the code that is giving problems is the one related to detect the variables in order to create the subscriptions in the CB. The following function is used to parse the expression and get the variables:

https://github.com/telefonicaid/iotagent-node-lib/blob/7b7c9506bec8db6a29ac8ef3a193e969f09334e6/lib/plugins/bidirectionalData.js#L73-L92

fgalan:
I think we should chose between two options:

1. Adapt bidirectonality so it works with no-legacy lenguace (JEXL), whatever it means... :)
2. Keep bidirectionality plugin as it is but tell in documentation about this gap/lack in support

mapedraza:
> I think we should chose between two options:
> 
> 1. Adapt bidirectonality plugin so it works with no-legacy lenguace (JEXL), whatever it means... :)
> 2. Keep bidirectionality plugin as it is but tell in documentation about this gap/lack in support

2nd alternative is not possible if we want completely remove legacy expressions. From my point of view, there are 3 alternatives if we want to completely remove legacy expressions from codebase:

1. Adapt bidirectonality plugin so it works with no-legacy language (JEXL), including code to automatically detect variables contained in JEXL expression. This is harder compared to legacy language since variables were preceded with `@`
2. Adapt bidirectonality plugin so it works JEXL, including new fields for reverse variables (so IoTA will be able to use these fields to create the subscription)
3. Remove Legacy expression language and bidirectionality.

fgalan:
After some discussion we are going to go with

> Remove Legacy expression language and bidirectionality.

except if somebody currently using ittells otherwise before May 9th, 2023 (hard deadline). As far as we know, nobody is using bidirectionally plugin.

Anyway, if after the removal some use case gets broken, then we can re-introduce bidirectionally with a clean implementation using JEXL (same strategy than the one used with the "magic cast" removed in 3.0.0).

mapedraza:
Based on history messages on IoTA repos, this could impact:
@bauer
@bobeal 
@martin-p-bauer
@RupamNEC
@ccndukwe
@Blobonat
@MayankSingh-NEC
@Sunny-NEC

config.js
```diff
@@ -30,11 +30,11 @@ var config = {
     server: {
```
fgalan, line 30: CNR entry regarding the changes in this PR is missing.

fgalan: I understand mentioning not only the removal of the legacy expression language, but also the fix of the bidirectional plugin to work with JEXL.

fgalan: I understand, it should mention issue #1340

doc/deprecated.md
```diff
@@ -21,6 +21,7 @@ A list of deprecated features and the version in which they were deprecated foll
 -   Support to NGSI-LD v1.3 in iotagent-node-lib 2.25.0 (finally removed in 2.26.0)
 -   Support groups (provision) statically defined by configuration
 -   Support to in-memory registry (i.e.`deviceRegistry.type=memory`)
+-   Support to legacy expressions
```
fgalan, line 24: Actually, this is not a deprecation case (in the sense it is still supported although not evolved, documented, etc.) but a removal case. Instead of adding the bullet here, maybe it makes sense to incluide it in the table at the end of the document.

fgalan:
```suggestion
-   Support to legacy expressions (finally removed in 3.1.0)
```

mapedraza: Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1339/commits/cd4a4862ad16300ba95d0dd34e78bc4d47762b30

mapedraza: added: 
> (finally removed in 3.2.0)

Because 3.1.0 is still supporting Legacy Expression

doc/deprecated.md
```diff
@@ -48,4 +49,5 @@ The following table provides information about the last iotagent-node-lib versio
 | Support to Node.js v8  | 2.12.0                                                | April 7th, 2020               |
 | Support to Node.js v10 | 2.15.0                                                | February 18th, 2021           |
 | Support to Node.js v12 | 2.24.0                                                | September 2nd, 2022           |
-| Support to NGSI-LD 1.3 | 2.25.0                                                | January 24th, 2023          |
+| Support to NGSI-LD 1.3 | 2.25.0                                                | January 24th, 2023            |
+| Support to Legacy Expressions | 3.1.0                                          | April 25th, 2023              |
```
fgalan, line 53:
```suggestion
| Support to Legacy Expressions | 3.0.0                                          | April 25th, 2023              |
```

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,5 @@
+- Add: JEXL support to bidirectinal plugin
+- Remove: remove legacy expression support (#1340)
```
fgalan, line 2:
```suggestion
- Remove: remove legacy expression support (removing IOTA_DEFAULT_EXPRESSION_LANGUAGE env var an associated defaultExpressionLanguage config.js setting) (#1340)

```

doc/deprecated.md
fgalan: I missleaded the version number. NTC

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,6 @@
+- Add: JEXL support to bidirectinal plugin
+- Remove: remove legacy expression support (removing IOTA_DEFAULT_EXPRESSION_LANGUAGE env var and associated defaultExpressionLanguage config.js setting) (#1340)
+
```
fgalan, line 3:
```suggestion
```

PR #1343, opened 2023-04-24
[HOTFIX] propage TimeInstant to all metatata when is provided ad measure 

AlvaroVega:
propage TimeInstant to all metatata when is provided ad measure  for al lcases

- [x] Upgrade tests


AlvaroVega:
Decision tree:

fgalan (approved):
LGTM

PR #1348, opened 2023-04-26
Re-enable Custom Update Plugins

jason-fox:
When upgrading from 2.x to 3.x I noticed that custom middlewares were no longer being invoked on updates. This PR restores that functionality.

- Add call to chain invoke plugins prior to update
- Add unit tests for `update()` and `query()` to check plugins are called.


jason-fox:
Note that the *IoT Agent for ISOXML*  https://github.com/fiware/iotagent-isoxml cannot be upgraded to use 3.x IoT Agent node lib until this issue has been fixed and a release made.

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -352,7 +350,7 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
         attributes,
         typeInformation
     );
-    let payload = {
+    const payload = {
```
mapedraza, line 353: Before reviewing the PR, I kindly ask you if you could "rollback" the lint changes in this file?

That would remove lot of noise and reduce the scope of the PR

Thanks

CHANGES_NEXT_RELEASE
```diff
@@ -1,8 +1,9 @@
...
-- Fix: allow use static attributes from group in entityNameExp of group
-- Fix: disable device attribute entity_name validation using pattern
+- Fix: reenable invocation of custom plugins on update (#1348)
+- Add: JEXL support to bidirectinal plugin
+- Remove: remove legacy expression support (removing IOTA_DEFAULT_EXPRESSION_LANGUAGE env var and associated defaultExpressionLanguage config.js setting) (#1340)
+- Fix: avoid usage of newlines in logs (#1342)
+- Fix: explicitAttributes true case: should progress just active attributes which receives measures and all active attributes with expressions
+- Fix: add missed active attributes (with expressions) for ctxt for evaluate explicitAttrs expression (#1351)
+- Fix: allow use static attributes from group in entityNameExp of group
+- Fix: disable device attribute entity_name validation using pattern
```
mapedraza, line 8: Group CNR categories

```suggestion
- Add: JEXL support to bidirectinal plugin
- Remove: remove legacy expression support (removing IOTA_DEFAULT_EXPRESSION_LANGUAGE env var and associated defaultExpressionLanguage config.js setting) (#1340)
- Fix: avoid usage of newlines in logs (#1342)
- Fix: explicitAttributes true case: should progress just active attributes which receives measures and all active attributes with expressions
- Fix: add missed active attributes (with expressions) for ctxt for evaluate explicitAttrs expression (#1351)
- Fix: allow use static attributes from group in entityNameExp of group
- Fix: disable device attribute entity_name validation using pattern
- Fix: reenable invocation of custom plugins on update (#1348)
```

fgalan (approved):
LGTM

PR #1355, opened 2023-04-28
Fix/explicit attributes true: just measures in active attrs and active attrs with expressions

AlvaroVega:
- [x] Add more tests about explicitAttributes: true

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,2 +1,1 @@
-
-
+- Fix explicitAttributes true case: should progress just active attributes which receives measures and all active attributes with expressions
--- a/lib/services/ngsi/entities-NGSI-v2.js
+++ b/lib/services/ngsi/entities-NGSI-v2.js
@@ -524,15 +524,18 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
                     selectedAttrs.push(attr.name);
                     selectedAttrs.push(attr.object_id);
                 } else {
-                    // check if attr is receiving a measure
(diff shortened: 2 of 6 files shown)


mapedraza (approved):
LGTM

CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
+- Fix: explicitAttributes true case: should progress just active attributes which receives measures and all active attributes with expressions
```
fgalan, line 1: This behaviour match the one described in documentation?

mapedraza: According to this:

https://github.com/telefonicaid/iotagent-node-lib/blob/ea48c800eaca591441a70be56b34d61b93b93dc9/doc/advanced-topics.md?plain=1#L301-L309

> *the IoTAgent rejects the  measure elements that are not defined in the mappings of device or group provision*

From my perspective, yes, it impact on it

AlvaroVega: https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/advanced-topics.md#explicitly-defined-attributes-explicitattrs

Case 2:

"explicitAttrs": true

just measures defined in active, static (plus conditionally TimeInstant) will be propagated to NGSI interface.

mapedraza: So, no further changes required in the documentation

fgalan: NTC

fgalan (approved):
LGTM

PR #1358, opened 2023-05-04
Fix/device attribute entity_name validation

AlvaroVega:
An attribute entity_name like `Device + ':' + Type` where rejected


fgalan (approved):
LGTM

PR #1360, opened 2023-05-04
Fix/log level tp info no context available for expr warn

AlvaroVega:
Warn level was introducing panic an so many messages in that production run level.
Three is nothing bad in "no context available for apply expression" since just expression is not evaluated


fgalan (approved):
LGTM

PR #1361, opened 2023-05-04
Fix bidirectional data

mapedraza:


lib/plugins/bidirectionalData.js
```diff
@@ -75,9 +75,7 @@ function extractVariables(item) {
```
fgalan, line 75: CNR entry missing?

mapedraza: The PR destination is another branch (the one that removes Legacy support). Final CNR entry should be included there with all the details

fgalan: Ok. NTC

fgalan (approved):
LGTM

PR #1362, opened 2023-05-04
add timeinstant to explicitAttrs true case

AlvaroVega:
Continuation of PR https://github.com/telefonicaid/iotagent-node-lib/pull/1355
Adds TimeInstant as is expected by https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/advanced-topics.md#explicitly-defined-attributes-explicitattrs
Case 2:

"explicitAttrs": true

just measures defined in active, static (plus conditionally TimeInstant) will be propagated to NGSI interface.

Allow refer to measure by attr name or attr object_id

### DIFF
--- a/lib/services/ngsi/entities-NGSI-v2.js
+++ b/lib/services/ngsi/entities-NGSI-v2.js
@@ -541,6 +541,8 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
         } else if (typeInformation.explicitAttrs && typeof typeInformation.explicitAttrs === 'boolean') {
             // explicitAtts is true => Add just measures which are defined in active attributes
             // and active attributes with expressions
+            // and TimeInstant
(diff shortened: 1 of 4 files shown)


lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -541,6 +541,8 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
         } else if (typeInformation.explicitAttrs && typeof typeInformation.explicitAttrs === 'boolean') {
```
fgalan, line 541: This PR needs to modify CNR? Or it is already covered by an existing entry?

AlvaroVega: No CNR is expected to be modified with this PR

fgalan: NTC

Issue #1353, opened 2023-04-27
 add skipValue to each attribute with expression


AlvaroVega:
SImplify explictAttrs to be only boolean ?
add to each attribute with expression a skipValue to allow not propagate attribute is as a result of expression result is skipValue

This way all logic from explicAttrs when is an expression will be translated into each attribute

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1364

PR #1364, opened 2023-05-08
Task/skip value for attr expression

AlvaroVega:
Issue https://github.com/telefonicaid/iotagent-node-lib/issues/1353

But this PR do not remove any explicitAttrs usage

- [x] Add doc about skip_value


doc/api.md
```diff
@@ -180,6 +180,7 @@ Some transformation plugins also allow the use of the following optional fields:
 -   **expression**: indicates that the value of the target attribute will not be the plain value or the measurement, but
     an expression based on a combination of the reported values. See the
     [Expression Language definition](#expression-language-support) for details
+-   **skip_value**: indicates the value to discard attribute of be sent ti CB when as a result of `expression`.
```
fgalan, line 183: skip_value (like entity_name) or skipValue (like explicitAttrs)? I would say that all the fields added recently follow the camelCase style, but I'm not fully sure.

Anyway, I'm afraid the mixture of naming styles we have in the API would be difficult to solve due to backward compatibility :(

AlvaroVega: skip_value like entity_name, since is at the same level

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -742,6 +742,17 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
                 try {
                     if (expressionPlugin.contextAvailable(attr.expression, ctxt, typeInformation)) {
                         res = expressionPlugin.applyExpression(attr.expression, ctxt, typeInformation);
+                        if (attr.skip_value !== undefined && res === attr.skip_value) {
```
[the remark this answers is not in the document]
AlvaroVega: CNR updated in https://github.com/telefonicaid/iotagent-node-lib/pull/1364/commits/2706475dde91b5f9a29cd2c3ecc199ce5d5e76c1

AlvaroVega:
After talk with @mrutid defualt skip_value should be null if nothing is defined

doc/api.md
```diff
@@ -180,6 +180,8 @@ Some transformation plugins also allow the use of the following optional fields:
 -   **expression**: indicates that the value of the target attribute will not be the plain value or the measurement, but
     an expression based on a combination of the reported values. See the
     [Expression Language definition](#expression-language-support) for details
+-   **skipValue**: indicates the value to discard attribute of be sent ti CB when as a result of `expression`. By
+    default if skipValue is not defined then is considered as `null`.
```
fgalan, line 184:
```suggestion
-   **skipValue**: indicates the value so if the measure value (or the result of applying `expression` to it, in the case `expression` is used) match `skipValue` then the attribute corresponding to the measure is not sent to CB. By
    default if `skipValue` is not defined then is considered as `null` (i.e. if the measure value or the evaluation of the expression results in `null` then attribute is not sent to CB).
```

doc/api.md
```diff
@@ -180,6 +180,10 @@ Some transformation plugins also allow the use of the following optional fields:
 -   **expression**: indicates that the value of the target attribute will not be the plain value or the measurement, but
     an expression based on a combination of the reported values. See the
     [Expression Language definition](#expression-language-support) for details
+-   **skipValue**: indicates the result of applying `expression` to a measure is equal to the value of `skipValue` then
+    the attribute corresponding to the measure is not sent to CB. By default if `skipValue` is not defined then is
+    considered as `null` (i.e. if the result of apply `expression` results in `null` then corresponding attribute is not
+    sent to CB).
```
fgalan, line 186:
```suggestion
-   **skipValue**: indicates that if the result of applying `expression` to a measure is equal to the value of `skipValue` then
    the attribute corresponding to the measure is not sent to CB. By default if `skipValue` is not defined then is
    considered as `null` (i.e. if the result of apply `expression` results in `null` then corresponding attribute is not
    sent to CB). It is only used if `expression` is provided (otherwise is ignored).
```

fgalan (approved):
LGTM

PR #1382, opened 2023-06-14
Improve Append Mode doc

mapedraza:
Comes from: 
https://github.com/telefonicaid/iotagent-ul/issues/216#issuecomment-1590810453

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -448,6 +448,19 @@ This is a flag that can be enabled by activating the parameter `appendMode` in t
 activated, the update requests to the Context Broker will be performed always with APPEND type, instead of the default
 UPDATE. This have implications in the use of attributes with Context Providers, so this flag should be used with care.
 
+When running the agent using `appendMode=false`, if sending measures that are not included in the config group (as 
+active measures), the IoT Agent should return a `422 Unprocessable Entity` code with the following message:
+
+```json
+{
+    "name": "ENTITY_GENERIC_ERROR",
+    "message": "Error accesing entity data for device: deviceType:dev of type: deviceType"
+}
+```
+
+Additionaly, The agent should have created the device and the corresponding entity in the broker if `autoprovision==true` 
+(default behaviour).
+
 ### Differences between `autoprovision`, `explicitAttrs` and `appendMode`
 
 Since those configuration parameters are quite similar, this section is intended to clarify the relation between them.
--- a/doc/api.md
+++ b/doc/api.md
@@ -458,7 +458,7 @@ active measures), the IoT Agent returns a `422 Unprocessable Entity` code with t
 }
 ```
 
-Additionaly, the agent creates the device and the corresponding entity in the broker if `autoprovision==true` 
+Additionally, the agent creates the device and the corresponding entity in the broker if `autoprovision==true` 
 (default behaviour).
 
 ### Differences between `autoprovision`, `explicitAttrs` and `appendMode`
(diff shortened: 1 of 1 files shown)


doc/api.md
```diff
@@ -448,6 +448,19 @@ This is a flag that can be enabled by activating the parameter `appendMode` in t
 activated, the update requests to the Context Broker will be performed always with APPEND type, instead of the default
 UPDATE. This have implications in the use of attributes with Context Providers, so this flag should be used with care.
 
+When running the agent using `appendMode=false`, if sending measures that are not included in the config group (as 
+active measures), the IoT Agent should return a `422 Unprocessable Entity` code with the following message:
```
fgalan, line 452:
```suggestion
active measures), the IoT Agent returns a `422 Unprocessable Entity` code with the following message:
```

As I think is a description of functional behaviour and optionality is not possible here (I could be wrong).

doc/api.md
```diff
@@ -448,6 +448,19 @@ This is a flag that can be enabled by activating the parameter `appendMode` in t
...
+
+```json
+{
+    "name": "ENTITY_GENERIC_ERROR",
+    "message": "Error accesing entity data for device: deviceType:dev of type: deviceType"
+}
+```
+
+Additionaly, The agent should have created the device and the corresponding entity in the broker if `autoprovision==true`
```
fgalan, line 461:
```suggestion
Additionaly, the agent creates the device and the corresponding entity in the broker if `autoprovision==true` 
```

doc/api.md
```diff
@@ -448,6 +448,19 @@ This is a flag that can be enabled by activating the parameter `appendMode` in t
...
+active measures), the IoT Agent returns a `422 Unprocessable Entity` code with the following message:
+
+```json
+{
+    "name": "ENTITY_GENERIC_ERROR",
+    "message": "Error accesing entity data for device: deviceType:dev of type: deviceType"
+}
+```
+
+Additionaly, the agent creates the device and the corresponding entity in the broker if `autoprovision==true`
```
mapedraza, line 461:
```suggestion
Additionally, the agent creates the device and the corresponding entity in the broker if `autoprovision==true` 
```

fgalan:
@Keshav-NEC it would be great if you could have a look to this PR and tell us if it solves https://github.com/telefonicaid/iotagent-ul/issues/216#issuecomment-1590810453

Keshav-NEC:
Hi @fgalan, it would be great if someone will share the file path where I have to do changes in append mode.

AlvaroVega:
> it would be great if someone will share the file path where I have to do changes in append mode.

You just need to define in docker env var   - IOTA_APPEND_MODE=true of in config.json add  key config: appendMode = true

AlvaroVega (approved):
LGTM

fgalan:
Based in @AlvaroVega 's LGTM, this PR is going to be closed. @Keshav-NEC of course if new feedback comes after your test a new PR could be raised to improve documentation.

AlvaroVega:
@mapedraza @fgalan My question here is if is really needed the option about appendMode = false, when we need it ? Why don't assume always is appendMode true an simplify everything?

mapedraza:
The idea behind having appendMode=false is to not allow the IoT Agent creates new attributes on entities on the context broker side (similar to autoprovision but not the same) . I do not know if there is a practical scenario in which this is interesting. Probably is a good idea to change the default behavior to appendMode=true, at least, until further changes.

fgalan:
Maybe related discussion: https://github.com/telefonicaid/iotagent-node-lib/pull/1111#issuecomment-1589279315

Issue #1377, opened 2023-06-01
static attributes (and other configuration flags)  in groups and autoprovision devices


AlvaroVega:
Static attributes in groups are copied to autoprovision device data and before modifications of attribute static are not progression to entity.

AlvaroVega:
The reason of thtat is static_attributes are in this array:

https://github.com/telefonicaid/iotagent-json/blob/48a9fbc071c5fec05fe9241eb18d8509c800a637/lib/iotaUtils.js#L239

AlvaroVega:
Other configuration flags like timestamp, expressionLanguage, etc are just copied form group to new autoprovisioned device and futher modifications of group are not propaged to devices

AlvaroVega:
> Static attributes in groups are copied to autoprovision device data and before modifications of attribute static are not progression to entity.

When a static attribute of a groups is modified (maybe is just not so static) then this value is not going to device entities after a new measure arrives to device; until then attribute has previous value in entity device.

AlvaroVega:
We are going to keep this issue open until clarify issue bout static_attrs with auto provisioned devices

fgalan:
After the merging of PR https://github.com/telefonicaid/iotagent-node-lib/pull/1396, should this issue be closed?

PR #1396, opened 2023-06-27
do not include static, lazy and commands in complete registrations

AlvaroVega:
Related with issue https://github.com/telefonicaid/iotagent-node-lib/issues/1377

To avoid duplicate  static, lazy and commands from groups into device

They were added by  https://github.com/telefonicaid/iotagent-node-lib/commit/c7e663c7e375ff616c78a973403a7f340aed05a2

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,4 +1,4 @@
-- Fix: explicitAttrs of device was tained even if not defined
+- Fix: explicitAttrs of device was tainted even if not defined
 - Fix: do not include static, lazy and commands in complete registrations to avoid duplicate them in device (#1377) 
 - Add: allow update timestamp and other config fields of device
 - Fix: use 'options=upsert' when update ngsiv2 CB entities and appendMode is enabled (#956)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,3 +1,4 @@
+- Fix: explicitAttrs of device was tained even if not defined
 - Fix: do not include static, lazy and commands in complete registrations to avoid duplicate them in device (#1377) 
 - Add: allow update timestamp and other config fields of device
 - Fix: use 'options=upsert' when update ngsiv2 CB entities and appendMode is enabled (#956)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,3 +1,4 @@
+- Fix: do not include static, lazy and commands in complete registrations to avoid duplicate them in device (#1377) 
 - Add: allow update timestamp and other config fields of device
 - Fix: use 'options=upsert' when update ngsiv2 CB entities and appendMode is enabled (#956)
 - Fix: do not propagate group config (timestamp and explicitAttrs) to autoprovisioned devices (at database level) (#1377)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,5 +1,5 @@
 - Fix: explicitAttrs of device was tainted even if not defined
-- Fix: do not include static, lazy and commands in complete registrations to avoid duplicate them in device (#1377) 
+- Fix: do not include static, lazy and commands from group to device to avoid duplicate them in device (#1377) 
 - Add: allow update timestamp and other config fields of device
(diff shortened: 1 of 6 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,5 @@
+- Fix: explicitAttrs of device was tained even if not defined
+- Fix: do not include static, lazy and commands in complete registrations to avoid duplicate them in device (#1377)
```
fgalan, line 2: Does it requires modifications to documentation? Or maybe documentation doesn't describe the "internals" of the registration process?

@mapedraza what do you think?

AlvaroVega: Maybe doc should be update, since this was reported as bug but there were tests about that previous behaviour

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,5 @@
+- Fix: explicitAttrs of device was tainted even if not defined
+- Fix: do not include static, lazy and commands from group to device to avoid duplicate them in device (#1377)
```
fgalan, line 2: I think the following paragraph in documentation in doc/api.md

> If devices are not pre-registered, they will be automatically created when a measure arrives to the IoT Agent - this
process is known as autoprovisioning. The IoT Agent will create an empty device with the group `apiKey` and `type` - the associated document created in database doesn't include config group parameters (`timestamp` and `explicitAttrs` in particular). The IoT Agent will also create the entity in the Context Broker if it does not exist yet.

should mention the elements this PRs add to the exclusion, I mean change this part to:

> (in particular, `timestamp`, `explicitAttrs`, static and lazy attributes and commands)

CHANGES_NEXT_RELEASE
mapedraza: > Does it requires modifications to documentation? Or maybe documentation doesn't describe the "internals" of the registration process?
> 
> @mapedraza what do you think?

As described here, this is the expected behaviour, so, IMHO, there is nothing to change into the documentation
https://github.com/telefonicaid/iotagent-node-lib/blob/0ab531d30587d6d7415b770e0f9072fce3f910d0/doc/api.md?plain=1#L149-L151

CHANGES_NEXT_RELEASE
mapedraza: > should mention the elements this PRs add to the exclusion, I mean change this part to:
> 
> > (in particular, `timestamp`, `explicitAttrs`, static and lazy attributes and commands)

At this point, would also make sense to include `active` (or attributes in IoTA):

> (in particular, `timestamp`, `explicitAttrs`, `active` ir `attributes`, `static` and `lazy` attributes and commands)

fgalan: Let's use @mapedraza 's version

AlvaroVega: Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1396/commits/e9af104cf73c7a707a6f00d50cde213545f39df6

doc/api.md
```diff
@@ -143,8 +143,9 @@ described in the [Device datamodel](#device-datamodel) section.
 
 If devices are not pre-registered, they will be automatically created when a measure arrives to the IoT Agent - this
 process is known as autoprovisioning. The IoT Agent will create an empty device with the group `apiKey` and `type` - the
-associated document created in database doesn't include config group parameters (`timestamp` and `explicitAttrs` in
-particular). The IoT Agent will also create the entity in the Context Broker if it does not exist yet.
+associated document created in database doesn't include config group parameters (in particular, timestamp, explicitAttrs,
+active or attributes, static and lazy attributes and commands). The IoT Agent will also create the entity in the
```
fgalan, line 147: If they refer to API fields, I think is better to use code-style for them, this way:

```suggestion
associated document created in database doesn't include config group parameters (in particular, `timestamp`, `explicitAttrs`,
`active` or `attributes`, `static` and `lazy` attributes and commands). The IoT Agent will also create the entity in the
```

fgalan (approved):
LGTM

mapedraza (approved):
LGTM

mapedraza:
Big coverage drop due to NGSI-LD test commented

fgalan:
> Big coverage drop due to NGSI-LD test commented

Let's hope that when these tests get re-enabled we see a similar raise in that PR :)

Issue #1384, opened 2023-06-21
Remove expressionLanguage fields from templates


mapedraza:
After removing expressionLanguage, should those fields be removed from the following files?

https://github.com/telefonicaid/iotagent-node-lib/blob/2f87e77240a722c2dcd9d0db9f2b064db9343afc/lib/templates/createDevice.json#L39-L42

https://github.com/telefonicaid/iotagent-node-lib/blob/2f87e77240a722c2dcd9d0db9f2b064db9343afc/lib/templates/createDeviceLax.json#L39-L42

https://github.com/telefonicaid/iotagent-node-lib/blob/1e31766d876a7b0d720a08984ca82f27440a9e59/lib/templates/deviceGroup.json#L40-L43

https://github.com/telefonicaid/iotagent-node-lib/blob/1e31766d876a7b0d720a08984ca82f27440a9e59/lib/model/Device.js#L30-L57

https://github.com/telefonicaid/iotagent-node-lib/blob/1e31766d876a7b0d720a08984ca82f27440a9e59/lib/services/groups/groupRegistryMongoDB.js#L39-L62

CC: @AlvaroVega

AlvaroVega:
expressionLanguage has not being removed completely.
expressionLanguage is still part of model of Devices, Groups, an checked in some many places of the library:

https://github.com/telefonicaid/iotagent-node-lib/blob/1e31766d876a7b0d720a08984ca82f27440a9e59/lib/model/Device.js#L54
https://github.com/telefonicaid/iotagent-node-lib/blob/1e31766d876a7b0d720a08984ca82f27440a9e59/lib/services/devices/deviceService.js#L181

But anyway we can remove all references to expresionLanguage, even in model assuming that always is jexl

fgalan:
> expressionLanguage has not being removed completely.
> expressionLanguage is still part of model of Devices, Groups, an checked in some many places of the library:

In my opinion, it should be removed in these cases. It doesn't make sense to have a field which values can take only one possibility. That's equivalent to not having the field, which is a much simpler solution.

@mapedraza maybe you could propose a PR with such removel to see how it likes?

mapedraza:
Solved by https://github.com/telefonicaid/iotagent-node-lib/pull/1398

Issue #1340, opened 2023-04-18
Remove support for legacy expression parser


AlvaroVega:

fgalan:
PR #1339

PR #1398, opened 2023-06-30
fully remove expression language field

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/issues/1384

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,3 +1,4 @@
+- Remove expression language (#1384)
 - Add: allow update timestamp and other config fields of device
 - Fix: use 'options=upsert' when update ngsiv2 CB entities and appendMode is enabled (#956)
 - Fix: do not propagate group config (timestamp and explicitAttrs) to autoprovisioned devices (at database level) (#1377)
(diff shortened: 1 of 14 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Remove expression language (#1384)
```
fgalan, line 1: We already mentioned this in https://github.com/telefonicaid/iotagent-node-lib/releases/tag/3.2.0

> Remove: remove legacy expression support (removing IOTA_DEFAULT_EXPRESSION_LANGUAGE env var and associated defaultExpressionLanguage config.js setting) (https://github.com/telefonicaid/iotagent-node-lib/issues/1340)

So I'd suggest something like this:

```suggestion
- Hardening: remove all stuff related with legacy expressions (removed in 3.2.0 but some leftovers remains in the code) (#1384)
```

fgalan: Or maybe better:

```suggestion
- Remove expressionLanguage field (as it it not longer needed, due to only one expression language is supported, after the removal of legacy expressions in 3.2.0) (#1384)
```

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Remove expressionLanguage field (as it it not longer needed, due to only one expression language is supported, after the removal of legacy expressions in 3.2.0) (#1384)
```
```suggestion
- Remove: expressionLanguage field (as it it not longer needed, due to only one expression language is supported, after the removal of legacy expressions in 3.2.0) (#1384)
```

fgalan (approved):
LGTM

PR #1391, opened 2023-06-26
Do not propage group config to autoprovision device

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/1377

This way autorpovisioned devices with remain without some flags defined like timestamp, expressionLanguage (which will take in runtime from group configuration)

- [x] Similar PRs in iotagent-json and iotagent-ul to align with local  findOrCreate

PR #1401, opened 2023-07-04
Revert "Update deviceService.js"

AlvaroVega:
Reverts telefonicaid/iotagent-node-lib#1400

explicitAttrs from config only needs be deleted from findorCreate as was done in https://github.com/telefonicaid/iotagent-node-lib/pull/1391
But mergeWithConfig needs that all non defined flags in device were filled with group values


fgalan (approved):
LGTM

PR #1405, opened 2023-07-19
FIX: Legacy expression tool 

mapedraza:
This PR fix 2 issues related with the script exectuion
- Fix the query when using `--expressionlanguage jexl`. Previously, the query includes all the grouops that have the `expressionlanguage` field. This increases the number of documents found and it is not relevant since the default value is legacy. The new behavior for `jexl` mode is not including new documents to the query
- Before this PR, the script only replace the expressionvalue if exists. This is not helping when the devices or groups does not have that parameter defined (remember, the default value is legacy, and, if not present, would be legacy as well after the actuation). The new behaviour is just forcing to append `"expressionlanguage":"jexl"` to each document modified


scripts/legacy_expression_tool/README.md
```diff
@@ -82,24 +82,24 @@ python legacy_expression_tool.py \
...
+| `--database`           | The database name to replace the expressions                                                             | NA                           | Yes       |
+| `--collection`         | The collection name to replace the expressions                                                           | NA                           | Yes       |
+| `--translation`        | The translation dictionary file to replace the expressions                                               | `translation.json`           | No        |
+| `--debug`              | Enable debug mode                                                                                        | `False`                      | No        |
+| `--commit`             | Commit the changes to the database                                                                       | `False`                      | No        |
+| `--expressionlanguage` | What to do with the expression language field. Possibles values: `delete`, `ignore`, `jexl` or `jexlall` | `ignore`                     | No        |
```
fgalan, line 93:
```suggestion
| `--expressionlanguage` | What to do with the expression language field. Possibles values: `delete` (to delete the field), `ignore` (to leave the field untouched), `jexl` (...) or `jexlall` (...) | `ignore`                     | No        |
```

The `(...)` has to be filled. Not sure the difference between `jexl` and `jexlall` so I cannot complete myself...

scripts/legacy_expression_tool/README.md
```diff
@@ -204,3 +204,38 @@ python legacy_expression_tool.py \
...
+${@value*100/total}                                    1        1   2
+All                                                    1        1   2
+```
+
+### Known issues
+
+#### Execution with `expressionlanguage` set to `jexlall`
+
+When executing the script with `expressionlanguage` set to `jexlall`, the script will look for all the documents
+containing legacy expressions or the parameter `expressionLanguage`. This would change the number of documents found,
```
fgalan, line 228: What "containing legacy expression" mean? Should be detailed.

I understand that when `jexl` is used the statistics are accurate. If I'm right, then I'd suggest to explicitly say it here.

scripts/legacy_expression_tool/README.md
```diff
@@ -204,3 +204,38 @@ python legacy_expression_tool.py \
...
+and the statistics will be different.
+
+#### Replacing expression without setting jexl at group or device level
+
+When executing the script without setting `expressionlanguage` to `jexl` or `jexlall` (or `delete`), the script will not
+change the`expressionLanguage` field in the document. This means that the legacy expressions will be replaced, but the
```
fgalan, line 234:
```suggestion
When executing the script setting `--expressionlanguage` to `ignore` (or when `--expressionlanguage` is not used), the script will not
change the`expressionLanguage` field in the document. This means that the legacy expressions will be replaced, but the
```

I think is more clear that way.

scripts/legacy_expression_tool/README.md
mapedraza:
```suggestion
| `--expressionlanguage` | What to do with the expression language field. Possibles values: `delete` (to delete the field), `ignore` (to leave the field untouched), `jexl` (to force to jexl on all groups that match the conditions) or `jexlall` (that, additionally, includes all groups that have a different expression language value than jexl to the query) | `ignore`                     | No        |
```

scripts/legacy_expression_tool/README.md
```diff
@@ -204,3 +204,38 @@ python legacy_expression_tool.py \
...
+All                                                    1        1   2
+```
+
+### Known issues
+
+#### Execution with `expressionlanguage` set to `jexlall`
+
+When executing the script with `expressionlanguage` set to `jexlall`, the script will look for all the documents
+containing legacy expressions or the parameter `expressionLanguage`. This would change the number of documents found,
+and the statistics will be different.
```
mapedraza, line 229:
```suggestion
and the statistics will be different. Running the script with the option set to `jexl` would not have impact on  the statistics. 
```

scripts/legacy_expression_tool/README.md
mapedraza: Solved in https://github.com/telefonicaid/iotagent-node-lib/pull/1405/commits/acc754a5baf9076af420431aade4b6cd76e46ea1

scripts/legacy_expression_tool/README.md
fgalan:
```suggestion
| `--expressionlanguage` | What to do with the expression language field. Possibles values: `delete` (to delete the field), `ignore` (to leave the field untouched), `jexl` (to force to jexl on all groups/devices that match the conditions) or `jexlall` (like `jext` but additionally including all groups/devices that have a different expression language value than jexl to the query) | `ignore`                     | No        |
```

scripts/legacy_expression_tool/README.md
```diff
@@ -82,24 +82,24 @@ python legacy_expression_tool.py \
...
+| `--collection`         | The collection name to replace the expressions                                                           | NA                           | Yes       |
+| `--translation`        | The translation dictionary file to replace the expressions                                               | `translation.json`           | No        |
+| `--debug`              | Enable debug mode                                                                                        | `False`                      | No        |
+| `--commit`             | Commit the changes to the database                                                                       | `False`                      | No        |
+| `--expressionlanguage` | What to do with the expression language field. Possibles values: `delete` (to delete the field), `ignore` (to leave the field untouched), `jexl` (to force to jexl on all groups/devices that match the conditions) or `jexlall` (like `jexl` but additionally including all groups/devices that have `expressionLanguage` field to the query) | `ignore`                     | No        |
```
fgalan, line 93:
```suggestion
| `--expressionlanguage` | What to do with the expression language field. Possibles values: `delete` (to delete the `expressionLanguage` field), `ignore` (to leave the field untouched, **FIXME** don't use, it can leave documents inconsistent), `jexl` (to change expressions to JEXL and in addition to change `expressionLanguage` field to `jexl` when the group/device uses legacy in some of the expression-related field, no matter if the group/devices have previously the `expresionLanguage` field or not) or `jexlall` (like `jexl` option, but also including groups/devices that doesn't use JEXL i any expression-related field but have `expresionLanguage` field defined, no matter its value) | `ignore`                     | No        |
```

fgalan (approved):
LGTM (although a followup PR with additional doc fixes will come)

Issue #1407, opened 2023-07-21
Payload measures are lost after reconnection with Orion


pooja1pathak:
### Component

Feature enhancement

### Is your feature request related to a problem? Please describe

Issue is related to: https://github.com/telefonicaid/iotagent-json/issues/718

### Describe the solution you'd like

IOT Agent should retry to send data in case orion is not reachable.

Keshav-NEC:
Hi, I am working on it.

PR #1408, opened 2023-07-21
Fix_for_OrionCB Payload loss

Keshav-NEC:
Fix for issue #1407

### DIFF
--- a/lib/request-shim.js
+++ b/lib/request-shim.js
@@ -52,7 +52,7 @@ function getOptions(options) {
         searchParams: options.searchParams || options.qs,
         headers: options.headers,
         throwHttpErrors: options.throwHttpErrors || false,
-        retry: options.retry || 0,
+        retry: options.retry || 5,
         responseType: options.responseType || 'json'
     };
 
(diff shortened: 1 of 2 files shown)


Keshav-NEC:
Gentle Reminder!

lib/request-shim.js
```diff
@@ -52,7 +52,7 @@ function getOptions(options) {
         searchParams: options.searchParams || options.qs,
         headers: options.headers,
         throwHttpErrors: options.throwHttpErrors || false,
-        retry: options.retry || 0,
+        retry: options.retry || 5,
```
fgalan, line 55: Taken into account this `retry` parameter, I wonder if the loop based in `retry_count` is really needed. Could you include got library documentation about this `retry` option, please?

Keshav-NEC: Hi @fgalan , I include this retry count to make the default value less than httpOptions.retry so that the agent can retry 5 times before error.

>Same here in https://github.com/telefonicaid/iotagent-json/blob/ef782f8f89ea2d0769dc36ecf9d96c63ef82cc17/lib/bindings/AMQPBinding.js#L130
and numRetried is used.

fgalan: I mean, modifications from L98 to L130 wouldn't be necessary, as far as I understand.

Keshav-NEC: Hi @fgalan , I have changed the default value but the agent was not retrying to connect to orion CB when it was down. Thats why I have to code to make it retry 5 times and also through an error as it was initially doing but after retrying 5 times after a particular interval of time. Thanks for the reply.

fgalan: Do you mean that got library is not working properly? As far as I understand, the got library should deal with this.

https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md shows a pretty sophisticated configuration. I'd suggest to see how it works and use it.

Keshav-NEC: Hi @fgalan , I just want to say to make that work we need some value to make it work 5 times so that it comes out of loop after max retries as it is implemented also in 
>https://github.com/telefonicaid/iotagent-json/blob/ef782f8f89ea2d0769dc36ecf9d96c63ef82cc17/lib/bindings/AMQPBinding.js#L130

and line ```145```

mapedraza:
Did you test yourself if it works? I do not think so... Anyway, a tests verifying this new behaviour, together with the documentation are required.

mapedraza:
As @fgalan mentioned, this PR as it is right now is not the best way to implement CB retries, as the got driver implements retry options. The best is to wire those options (right now is disabled) 

Some test cases should be implemented, reflecting a failling connection and working after some retries. 

This should be configured as per device group, since the CB is also configurable at device group.

Issue #1404, opened 2023-07-13
Timeinstant config parameter is ignored


mapedraza:
### IoT Agent Node Lib version the issue has been seen with

3.2.0

### Bound or port used (API interaction)

Southbound (Devices data API), Northbound (Provision API and NGSI Interactions)

### NGSI version

NGSIv2

### Are you running a container?

Yes, I am using a contaner (Docker, Kubernetes...)

### Image type

normal

### Expected behaviour you didn't see

When using `timeInstant`parameter under a group, it is not working as expected. 6 possible cases analyzed

| `timeInstant` value | measure contains timeInstant | actual behaviour  | expectation |
| -- | -- | -- | -- |
| `true` | Yes | TimeInstant and metadata updated with measure value | works well - OK |
| `true` | No | TimeInstant and metadata updated with server timestamp | works well - OK |
| `false` | Yes | TimeInstant and metadata updated with measure value | should appear as attribute but not as metadata - NOK |
| `false` | No | TimeInstant and metadata updated with server timestamp | should not appear neither as attribute not as metadata - NOK |
| Not defined | Yes | TimeInstant and metadata updated with measure value | works well - OK |
| Not defined | No | TimeInstant and metadata updated with server timestamp | works well - OK |

### Configs

```yaml
environment:
    - "IOTA_CB_HOST=orion"
    - "IOTA_CB_PORT=1026"
    - "IOTA_NORTH_PORT=4041"
    - "IOTA_REGISTRY_TYPE=mongodb"
    - "IOTA_MONGO_HOST=mongodb"
    - "IOTA_MONGO_PORT=27017"
    - "IOTA_MONGO_DB=iotagent-json"
    - "IOTA_HTTP_PORT=7896"
    - "IOTA_PROVIDER_URL=http://iot-agent:4041"
```

mapedraza:
Additionally, test does not implements all the cases:
https://github.com/telefonicaid/iotagent-node-lib/blob/master/test/unit/ngsiv2/plugins/timestamp-processing-plugin_test.js

AlvaroVega:
Maybe related https://github.com/telefonicaid/iotagent-node-lib/issues/1377 ?

AlvaroVega:
timeInstant value | measure contains timeInstant | actual behaviour | expectation
-- | -- | -- | --
false | Yes | TimeInstant and metadata updated with measure value | should appear as attribute but not as metadata - NOK
-- | -- | -- | --

[...]

PR #1415, opened 2023-08-23
Add `timestamp=false` tests

mapedraza:
This PR add 2 new tests when `timestamp=false`
- Without sending timeInstant attribute
- Sending TimeInstant attribute

The behaviour is the expected

This PR includes tests missing in #1404


fgalan (approved):
LGTM

Issue #1426, opened 2023-09-05
Wrong device use when several devices with same id but different apikey and type are used


AlvaroVega:
### IoT Agent Node Lib version the issue has been seen with

3.2.0

### Bound or port used (API interaction)

Southbound (Devices data API)

### NGSI version

NGSIv2

### Are you running a container?

Yes, I am using a contaner (Docker, Kubernetes...)

### Image type

normal

### Expected behaviour you didn't see

is not using right device

### Unexpected behaviour you saw

use the right device

### Steps to reproduce the problem

Create both: APIKEY-A/Type-A and APIKEY-B/Type-B groups

Sending measure to : APIKEY-A ID-0 (JEXL Entity_id=ID) Type-A -->  creates  entity ID-0 of Type-A  -> OK

Sending measure to: APIKEY-B ID-0 (JEXL Entity_id=ID) Type-B -->  creates attributes in entity  ID-0 of Type-A   -> NO OK

Sending measure: APIKEY-B ID-1 (JEXL Entity_id=ID) Type-B -->  creates entity ID-1 of Type-B -> OK

### Configs

```yaml
environment:
    - "IOTA_CB_HOST=orion"
    - "IOTA_CB_PORT=1026"
    - "IOTA_NORTH_PORT=4041"
    - "IOTA_REGISTRY_TYPE=mongodb"
    - "IOTA_MONGO_HOST=mongodb"
    - "IOTA_MONGO_PORT=27017"
    - "IOTA_MONGO_DB=iotagent-json"
    - "IOTA_HTTP_PORT=7896"
    - "IOTA_PROVIDER_URL=http://iot-agent:4041"
```

### Log output

```sh
NO logs
```

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1438

PR #1427, opened 2023-09-05
try to use apikey to find device at first attempt

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/1426

Type of group and device could be different (specially when device is provisioned explicitly)

This fix implies that all provisioned devices has a valid apikey


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
-- Fix: ensure device apikey in already provisioned device (#1430)
-- Upgrade mongodb dev dep from 4.7.0 to 4.17.0
-- Upgrade mongoose dep from 5.13.14 to 5.13.20
+- Try to use apikey from measure/group to find device in first attempt (#1426)
```
```suggestion
- Fix: try to use apikey from measure/group to find device in first attempt (#1426)
```

PR #1434, opened 2023-09-06
[WIP] FIX change to mongo 7.0 in regression tests

fgalan:


fgalan:
We are going to skip 7.0 and go directly to 8.0. Thus, this PR is no longer useful and it is closed.

Issue #1440, opened 2023-09-11
Not evaluable expression should be treated as null when using skipValue


mapedraza:
### IoT Agent Node Lib version the issue has been seen with

Latest (3.3.0-next)

### Bound or port used (API interaction)

Other

### NGSI version

NGSIv2

### Are you running a container?

Yes, I am using a contaner (Docker, Kubernetes...)

### Image type

None

### Expected behaviour you didn't see

When an active attribute expression involving a measurement (`"expression":"t*2"`)is evaluated, if missing parameter (I.E: when a device sends `{"m":3}` the resulting entity attribute is `null`. This is the expected.

When defining active attribute expression as constant `null` (`"expression":null`), the resulting entity attribute is `null`, and `"skipValue":null`, the attribute is not updated by the IoTA. This is the expected value

When, for a given active attribute, with `"skipValue":null` and an expression `"expression":"t*2"`, a measurement without that parameter (`{"m":3}`) arrives, the attribute is written with a `null` value. This is **not** the expected result. It is expected that IoTA does not update the value. 

There is a workaround for this, in that case defining the expression like `t?t*2:null`, which will force to null the value. 

Probably, this is because internally, the result of an expression that does not have all the parameters available is an error that, in a subsequent stage, is transformed to a null. And the comparison with skipValue, in order to avoid persisting it into the CB is done in between.

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/api.md
 By default if skipValue is not defined then is considered as null 

So this case

`When an active attribute expression involving a measurement ("expression":"t*2")is evaluated, if missing parameter (I.E: when a device sends {"m":3} the resulting entity attribute is null. This is the expected.`

is the same that
`When, for a given active attribute, with "skipValue":null and an expression "expression":"t*2", a measurement without that parameter ({"m":3}) arrives, the attribute is written with a null value. `

mrutid:
"When, for a given active attribute, with "skipValue":null and an expression "expression":"t*2", a measurement without that parameter ({"m":3}) arrives, the attribute is written with a null value. This is not the expected result. It is expected that IoTA should not update the value."

In JEXL null*2 == null no exception or error.  are we sure about this behaviour?

It's not the same than "nojson"|jsonparse (Error: SyntaxError: Unexpected end of JSON input) in which case we convey to return null value also.

AlvaroVega:
in this case: 
 ```javascript 
[...]

PR #1441, opened 2023-09-12
fix for non evaluable expressions which should not be progress to attrs values

AlvaroVega:
Fix for https://github.com/telefonicaid/iotagent-node-lib/issues/1440


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Fix for non evaluable expressions which should not be propaged to attrs values (#1440)
```
```suggestion
- Fix: for non evaluable expressions which should not be propagated to attrs values (#1440)
```

fgalan (approved):
LGTM

PR #1462, opened 2023-09-21
set useUnifiedTopology = true to remove warn

AlvaroVega:
remove warn about

(node:1) DeprecationWarning: current Server Discovery and Monitoring engine is deprecated, and will be removed in a future version. To use the new Server Discover and Monitoring engine, pass option { useUnifiedTopology: true } to the MongoClient constructor.

related with some warns reported in https://github.com/telefonicaid/iotagent-json/issues/645


AlvaroVega:
@fgalan make sense ?

fgalan:
> @fgalan make sense ?

I think removing that deprecation warning is a good idea.

What I'm not sure is if this should be mentioned in CNR. Probably yes.

fgalan (approved):
LGTM

PR #1466, opened 2023-09-21
Remove lava code

jason-fox:
Static analysis of the changes between 3.3.0 and 3.4.0 indicates that the `createInitialEntity` from NGSI-v2. This PR does the same for "LD "and "mixed". There is no change in the tests, but coverage should increase marginally


AlvaroVega (approved):
LGTM

PR #1471, opened 2023-09-27
Add Null expression tests

mapedraza:
https://github.com/telefonicaid/iotagent-node-lib/issues/1440#issuecomment-1736983813

When posting a measure `v` with value `Null` to a group with the following attributes:

**Expression**
```js
            active: [
                {
                    name: 'a',
                    type: 'Number',
                    expression: 'v'
                },
                {
                    name: 'b',
                    type: 'Number',
                    expression: 'v*3'
                },
                {
                    name: 'c',
                    type: 'Boolean',
                    expression: 'v==null'
                },
                {
                    name: 'd',
                    type: 'Text',
                    expression: "v?'no soy null':'soy null'"
                },
                {
                    name: 'e',
                    type: 'Text',
                    expression: "v==null?'soy null':'no soy null'"
                },
                {
                    name: 'f',
                    type: 'Text',
                    expression: "(v*3)==null?'soy null':'no soy null'"
                },
                {
                    name: 'g',
                    type: 'Boolean',
                    expression: 'v == undefined'
                }
            ]
```

**Expectations**
```js
                    v: {
                        value: null,
                        type: 'Number'
                    },
                    b: {
                        value: 0,
                        type: 'Number'
                    },
                    c: {
                        value: true,
                        type: 'Boolean'
                    },
                    d: {
                        value: 'soy null',
                        type: 'Text'
                    },
                    e: {
                        value: 'soy null',
                        type: 'Text'
                    },
                    f: {
                        value: 'no soy null',
                        type: 'Text'
                    },
                    g: {
                        value: true,
                        type: 'Boolean'
                    }
```

PR #1474, opened 2023-09-28
Fix: Remove nulls from context before evaluating expressions

mapedraza:
This PR delete null elements from context. This fix arithmetic problems due to JEXL library bug:
https://github.com/telefonicaid/iotagent-node-lib/issues/1440
https://github.com/TomFrost/Jexl/issues/133

Overpasses:
https://github.com/telefonicaid/iotagent-node-lib/pull/1471

### DIFF
--- a/lib/plugins/jexlParser.js
+++ b/lib/plugins/jexlParser.js
@@ -125,10 +125,22 @@ function extractContext(attributeList) {
 
 function applyExpression(expression, context, typeInformation) {
     logContext = fillService(logContext, typeInformation);
+    // Delete null values from context. Related:
+    // https://github.com/telefonicaid/iotagent-node-lib/issues/1440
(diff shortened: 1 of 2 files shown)


mapedraza:
From #1471 

> Conclusions:
>
>- case `b` (`v*3`)
>   - For Null values, is always failing (treating `Null` as 0)
>   - For NaN values, it works well, giving Null as response. 
>- case `f` (`(v*3)==Null?'soy null':'no soy null'`) is always failing
>- Using `explicitAttrs` acts as sending a measurement with null value 

This PR **solves** case `b`. Case `f` is still failing.

AlvaroVega:
Not sure if this approach is introducing a bit of overhead, and maybe just only at measure level approach could be enough https://github.com/telefonicaid/iotagent-node-lib/pull/1473

AlvaroVega (approved):

PR #1482, opened 2023-10-02
Add tests for ID and Type colisions

mapedraza:

PR #1483, opened 2023-10-02
FIX: Remove measures with attributes name 'id' or 'type' (hotfix 3.4.3)

AlvaroVega:
Issue https://github.com/telefonicaid/iotagent-node-lib/issues/1485

this should be tested 
- [x] with https://github.com/telefonicaid/iotagent-node-lib/pull/1482/
- [x]  curl -i -X POST 'http://localhost:7897/iot/json?i=disp1&k=idmaootoin375kl9iifz7472z' -d '{"m": 3, "id": "mierda", "type":"otro"}' -H 'content-type: application/json'

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,1 @@
-
+- FIX: remove attribute of measures with name `id` or `type` to avoid collisions
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,0 @@
-- FIX: remove attribute of measures with name `id` or `type` to avoid collisions
--- a/doc/api.md
+++ b/doc/api.md
@@ -152,7 +152,7 @@ parameters defined at device level in database, the parameters are inherit from 
 ## Entity attributes
 
 In the config group/device model there are four list of attributes with different purpose to configure how the
-information coming from the device is mapped to the Context Broker attributes:
+information coming from the device (measures) is mapped to the Context Broker attributes:
 
 -   **`attributes`**: Are measures that are pushed from the device to the IoT agent. This measure changes will be sent
     to the Context Broker as updateContext requests over the device entity. NGSI queries to the context broker will be
@@ -179,7 +179,8 @@ information coming from the device is mapped to the Context Broker attributes:
 All of them have the same syntax, a list of objects with the following attributes:
 
 -   **object_id** (optional): name of the attribute as coming from the device.
--   **name** (mandatory): ID of the attribute in the target entity in the Context Broker.
+-   **name** (mandatory): ID of the attribute in the target entity in the Context Broker. Note that `id`` and `type`` are 
+    not valid attribute names at Context Broker. Thus, although a id or type will not break the IOT Agent, they are silently ignored and never progress toward Context Broker entities.
 -   **type** (mandatory): name of the type of the attribute in the target entity.
 -   **metadata** (optional): additional static metadata for the attribute in the target entity. (e.g. `unitCode`)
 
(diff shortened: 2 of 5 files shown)


mapedraza:
> We should include at same point of the documentation (any suggestion @mapedraza ?) something like this:
> 
> > Note that `id` and `type` are not valid attribute names at Context Broker. Thus, although a `id` or `type` will not break the IOT Agent, they are silently ignored and never progress toward Context Broker entities.
> 
> or something with a similar meaning.

Probably, instead of deleting, renaming them to `_id` or `_type` could be useful in the future.

fgalan:
> Probably, instead of deleting, renaming them to `_id` or `_type` could be useful in the future

It depends on how strict we want to be with "illegal" measures. Deleting them will strongly disallow its use, while moving them to attributes `_id` and `_type` will somehow allow its usage.

In my opinion it's better to delete, but it's not a strong opinion.

mapedraza:
LGTM

mapedraza (approved):

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -340,6 +340,15 @@ function sendUpdateValueNgsi2(entityName, attributes, typeInformation, token, ca
         attributes,
         typeInformation
     );
+    // if any attribute name of a measure is 'id' or 'type' should be removed
```
fgalan, line 343:
```suggestion
    // if any measure has name 'id' or 'type' it should be removed
    // (as they cannot be progressed as attribute names, given that 'id' or 'type' are forbidden names for attributes in CB)
```

CHANGES_NEXT_RELEASE
```diff
@@ -1 +1 @@
-
+- FIX: remove attribute of measures with name `id` or `type` to avoid collisions
```
```suggestion
```

If this PR combines bump (note packages.json has been also modified) the CNR should be flushed. However, I'll keep this sentence for the future release notes in github.

doc/api.md
```diff
@@ -179,7 +179,9 @@ information coming from the device is mapped to the Context Broker attributes:
 All of them have the same syntax, a list of objects with the following attributes:
 
 -   **object_id** (optional): name of the attribute as coming from the device.
--   **name** (mandatory): ID of the attribute in the target entity in the Context Broker.
+-   **name** (mandatory): ID of the attribute in the target entity in the Context Broker. Note that `id`` and `type``
```
fgalan, line 182:
```suggestion
-   **name** (mandatory): ID of the attribute in the target entity in the Context Broker. Note that `id` and `type`
```

doc/api.md
```diff
@@ -179,7 +179,9 @@ information coming from the device is mapped to the Context Broker attributes:
 All of them have the same syntax, a list of objects with the following attributes:
 
 -   **object_id** (optional): name of the attribute as coming from the device.
--   **name** (mandatory): ID of the attribute in the target entity in the Context Broker.
+-   **name** (mandatory): ID of the attribute in the target entity in the Context Broker. Note that `id` and `type`
+    are not valid attribute names at Context Broker. Thus, although a id or type will not break the IOT Agent, they
```
fgalan, line 183:
```suggestion
    are not valid attribute names at Context Broker. Thus, although a measure named `id` or `type` will not break the IOT Agent, they 
```

doc/api.md
```diff
@@ -209,6 +211,10 @@ Additionally for commands (which are attributes of type `command`) the following
 -   **contentType**: `content-type` header used when send command by HTTP transport (ignored in other kinds of
     transports)
 
+Note that, when information comming from devices, this means measures, are not defined neither in the group, nor in the 
+device, the IoT agent will store that information into the destination entity using the same attribute name than the 
+measure name, unless `explicitAttrs` is defined. Measures `id` or `type` names are invaled, and will be ignored.
```
fgalan, line 216:
```suggestion
measure name, unless `explicitAttrs` is defined. Measures `id` or `type` names are invalid, and will be ignored. 
```

fgalan (approved):
LGTM

Issue #1476, opened 2023-10-02
id of entity (expanded by entityNameExp should be a string, even when result or expression is a number)


AlvaroVega:
### IoT Agent Node Lib version the issue has been seen with

3.3.0

### Bound or port used (API interaction)

Southbound (Devices data API)

### NGSI version

NGSIv2

### Are you running a container?

None

### Image type

None

### Expected behaviour you didn't see

Device 
```
{ "_id" : ObjectId("650d35422b1b15d5ca229e46"), "commands" : [ ], "staticAttributes" : [ ], "subscriptions" : [ ], "creationDate" : ISODate("2023-09-22T06:33:38.027Z"), "id" : "2699", "type" : "NoiseLevelObserved", "name" : "2699", "service" : "sc_vlci", "subservice" : "/sonometros", "apikey" : "e1sxijb6hvylgx94wm5k87jgw", "protocol" : "IoTA-JSON", "transport" : "HTTP", "timestamp" : false, "expressionLanguage" : "jexl", "__v" : 0 }
```

Group
```
{ "_id" : ObjectId("650bf6b42b1b153011f11629"), "attributes" : [ { "object_id" : "dateObservedFrom", "name" : "dateObservedFrom", "type" : "Timestamp" }, { "object_id" : "dateObservedTo", "name" : "dateObservedTo", "type" : "Timestamp" }, { "object_id" : "LAeq", "name" : "LAeq", "type" : "Float" } ], "resource" : "/iot/json", "apikey" : "e1sxijb6hvylgx94wm5k87jgw", "type" : "NoiseLevelObserved", "service" : "sc_vlci", "subservice" : "/sonometros", "description" : "OTP - VLCI - Sensores BSG - Ruido minutal", "timestamp" : false, "expressionLanguage" : "jexl", "entityNameExp" : "id", "__v" : 0 }
```

### Unexpected behaviour you saw

POST /v2/entities?options=upsert, request payload (237 bytes): {"id":2699,"type":"NoiseLevelObserved","dateObservedFrom":{"value":"2023-10-02T10:58:00+00:00","type":"Timestamp"},"dateObservedTo":{"value":"2023-10-02T10:59:00+00:00","type":"Timestamp"},"LAeq":{"value":63.721833278745,"type":"Float"}}, response code: 400

### Configs

```yaml
environment:
    - "IOTA_CB_HOST=orion"
    - "IOTA_CB_PORT=1026"
[...]

PR #1487, opened 2023-10-03
Fix/ensure entity id and type is string (fix for master)

AlvaroVega:
Issue #1476


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "iotagent-node-lib",
   "license": "AGPL-3.0-only",
   "description": "IoT Agent library to interface with NGSI Context Broker",
-  "version": "3.4.0-next",
+  "version": "3.4.3-next",
```
fgalan, line 5:
```suggestion
  "version": "3.4.0-next",
```

As the semantic for "x.y.0-next" use to be "the next .0 versfion after x.y.0" (typically, 3.5.0)

fgalan (approved):
LGTM

But I think it will be wise to wait a while for merging, while we check how hotfixes 3.4.x are working.

PR #1494, opened 2023-10-04
FIX improve nested expressions doc

fgalan:


doc/api.md
```diff
@@ -765,19 +765,61 @@ following to CB:
...
+    },
+    {
+        "name": "correctedLevel",
+        "type": "Number",
+        "expression": "level * 0.897"
+    },    
+]
+```
+
+It is not going to work. The first expression expects a `correctedLevel` which is neither a measure (remember the only measure sent by the device is named `level`) nor a previously calculated attribute. Thus, `correctedLevel` will end with a `null` value.
```
fgalan, line 806: Check this is correct. I'm pretty sure is `null` in this case but we have been struggling with this recently (`"null"`, `undefined`, `"undefined"` :) so please check.

doc/api.md
```diff
@@ -765,19 +765,61 @@ following to CB:
...
+
+In conclusion: **the order of attributes in the `attributes` arrays at provising time matters with regards to nested expression evaluation**.
+
+Let's consider the following example. It is an anti-pattern but it's quite illustrative on how ordering works:
 
 ```json
 "attributes": [
     {
-        "name": "attrA",
+        "name": "A",
```
fgalan, line 815: Better to use `A` and `B` as `attrX` is not correct for measure case (previously used in "When receiving a measure with the following values:").

Btw, I think there was a typo in second provision. It was using name:attrB-expression:attrB (now it uses name:A-exprresion:B)

NTC (just informative)

doc/api.md
mapedraza: Confirmed, it works like this

PR #1502, opened 2023-10-05
Update multientity-plugin_test.js and doc about explicitAttrs = true

AlvaroVega:


doc/api.md
```diff
@@ -417,7 +417,7 @@ Case 2:
 "explicitAttrs": true
 ```
 
-just measures defined in active, static (plus conditionally TimeInstant) will be propagated to NGSI interface.
+should progress just active attributes, static (plus conditionally TimeInstant)  which receives measures and all active attributes with expressions
```
mapedraza, line 420:
```suggestion
In this case, should only progress active and static attributes defined in the device or group provision (`TimeInstant`attrribute will be also included if enabled). In other words, having `"explicitAttrs":true` would prevent the IoTA creating attributes into the related entity  within the context broker from measures that are not explicitly defined in the device or group provision. 

Note that attributes defined in the provision that are not receiving a measure (or having a expression defined that is resulting `null`) will not progress (this means, the NGSI request to update the entity in the context broker is not going to include that attribute) unless `skipValue` is defined to other value than `null`
```

test/unit/ngsiv2/plugins/multientity-plugin_test.js
```diff
@@ -522,9 +522,10 @@ const iotAgentConfig = {
                     expression: 'z+1'
                 },
                 {
-                    name: 'nonexpectedAtt',
+                    name: 'alsoexpectedAtt',
                     type: 'number',
-                    expression: 'w+1'
+                    expression: 'w+1',
+                    skipValue: 'loquesea'
```
mapedraza, line 528: Why not also include notnexpectedAtt?
```suggestion
                    expression: 'w+1',
                    skipValue: 'loquesea'
                 },
                {
                    name: 'nonexpectedAtt'
                    type: 'number',
                    expression: 'w+1',
```

fgalan (approved):
LGTM

Issue #1526, opened 2023-11-13
MongoDB authSource is not using it


cblanco-libelium:
### IoT Agent Node Lib version the issue has been seen with

master

### Bound or port used (API interaction)

Other

### NGSI version

Other

### Are you running a container?

None

### Image type

None

### Expected behaviour you didn't see

When setting the IOTA_MONGO_AUTH_SOURCE it should be build the option on the connection string of mongo

### Unexpected behaviour you saw

It doesnt appear in the url, also tests are not correct because the expecting behaviour when using authSource is not concatening in the string connection

I'll post a PR now with the fix.

PR #1527, opened 2023-11-13
[Fix] Parse authSource variable into mongo url string

cblanco-libelium:
*Edit:* issue #1526 

    - Overriding when user:pass is present in extraArgs
    - Fix tests
    
    I see that authSource had an independant ENV variable and treatment along the code, but the truth it's that authSource should concat in the string as an extraArgs. 
    In my opinion the best way to allow backwards compatibility is to still parsing if comes in extraArgs, but overriding if the specific env variable is present. 
    Also authSource should only used if user:pass are present so i included in the if statemente


fgalan:
Thank you for your contribution!

Please add an entry to the CHANGES_NEXT_RELEASE file (which is used to build the Changelog of next version) to describe the fix in the PR. Something like this:

```
- Fix: mongodb.authSource / IOTA_MONGO_AUTH_SOURCE was not correctly supported (#1526)
```

AlvaroVega (approved):
LGTM

PR #1531, opened 2023-11-22
Skip explicitAttrs No Lib

mapedraza:
Modify tests to skip this cases in JSON agent untill the following issue is fixed:
https://github.com/telefonicaid/iotagent-json/issues/782

### DIFF
--- a/test/functional/testCases.js
+++ b/test/functional/testCases.js
@@ -2451,7 +2451,7 @@ const testCases = [
                 shouldName:
                     'D - WHEN no sending any defined case IT should store [static_a,static_b] attrs into Context Broker',
                 type: 'single',
-                skip: '!lib' // FIXME: https://github.com/telefonicaid/iotagent-json/issues/782
+                skip: '!lib', // FIXME: https://github.com/telefonicaid/iotagent-json/issues/782
                 measure: {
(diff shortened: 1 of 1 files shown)


test/functional/testCases.js
```diff
@@ -2424,6 +2424,7 @@ const testCases = [
             {
                 shouldName: 'C - WHEN sending only b IT should store only [{object_id:b}] attrs into Context Broker',
                 type: 'single',
+                skip: '!lib', // FIXME: https://github.com/telefonicaid/iotagent-json/issues/782
```
fgalan, line 2427: The usage of `lib` is documented in https://github.com/telefonicaid/iotagent-node-lib/blob/master/test/functional/README.md, but not the usage of `!lib`. Maybe that document should be improved to explain the `!lib` case.

mapedraza: You are right. Anyway, I think the best is to create a separated PR.

fgalan: Let's do it and merge both at the same time?

mapedraza: https://github.com/telefonicaid/iotagent-node-lib/pull/1532

PR #1540, opened 2023-12-12
store in device commands from group after CB registration

AlvaroVega:
Bug introduced in 3.3.0 by https://github.com/telefonicaid/iotagent-node-lib/pull/1396

commands from groups should be stored in device since registration is only possible against device entity.

Otherwise commands is stored in entity as normal string type attribute.

Moreover any command group modification requires device registration would be done again, and then implies device and entity should be deleted previously.


CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
+- FIX: store commands from Group at Device level at provision device time
```
```suggestion
- Fix: store commands from Group at Device level at provision device time
```

fgalan (approved):
LGTM

PR #1546, opened 2023-12-18
fix test

AlvaroVega:
Command is not sent to entity by POST /v2/entities;  command is stored in device and command is register into CB by iota in another request (POST TO /v2/registrations ) not covered by there test framework

logLevel in tests functional framework was doing a false positive resuilt test


fgalan (approved):
LGTM

PR #1562, opened 2024-01-29
Task/not loose invalid id type in measures

AlvaroVega:
related https://github.com/telefonicaid/iotagent-json/pull/804

PR #1567, opened 2024-01-30
[Doc] Add special measures and attributes names

mapedraza:
https://github.com/telefonicaid/iotagent-node-lib/pull/1562/files

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -149,6 +149,15 @@ This behavior allows that autoprovisioned parameters can freely established modi
 creation using the provisioning API. However, note that if a device (autoprovisioned or not) doesn't have these
 parameters defined at device level in database, the parameters are inherit from config group parameters.
 
+## Special measures and attributes names
+
+In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and 
+`measure_type`. The reason behind this is to avoid to collide with the original entity id and type, as mechanism 
+that enable store measure values from parameters called with the same name. It only applies to autoprovisioned
+attributes and is also available at JEXL context with the same name (`measure_id`or `measure_type`).
+
+In case of provisioning attributes using `id` or `type` as names, they are ignored. 
+
 ## Entity attributes
 
 In the config group/device model there are four list of attributes with different purpose to configure how the


doc/api.md
```diff
@@ -149,6 +150,15 @@ This behavior allows that autoprovisioned parameters can freely established modi
...
 parameters defined at device level in database, the parameters are inherit from config group parameters.
 
+## Special measures and attributes names
+
+In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and 
+`measure_type`. The reason behind this is to avoid to collide with the original entity id and type, as mechanism 
+that enable store measure values from parameters called with the same name. It only applies to autoprovisioned
+attributes and is also available at JEXL context with the same name (`measure_id`or `measure_type`).
+
+In case of provisioning attributes using `id` or `type` as names, they are ignored.
```
fgalan, line 160: I'm not sure if this is right...

Consider the following test:

https://github.com/telefonicaid/iotagent-node-lib/blob/5073e6ae1e6afe608925213b7f624c26e9248edf/test/unit/ngsiv2/ngsiService/active-devices-test.js#L166-L175

That's is a provisioning of `id` and `type`. Note is as an actual effect, as the type used there (`"text"`) is later use in the CB update as can be seen here:

https://github.com/telefonicaid/iotagent-node-lib/blob/5073e6ae1e6afe608925213b7f624c26e9248edf/test/unit/ngsiv2/ngsiService/active-devices-test.js#L936-L943

Maybe this can refrased as follows:

```suggestion
In case of provisioning attributes using `id` or `type` as names, the attribute type in `measure_id` and `measure`_type` are the ones used in that provision. However, provisioning `id` or `type` is considered and anti-pattern and your are encouraged to not doing so. 
```

@AlvaroVega could you confirm my understanding?

doc/api.md
```diff
@@ -149,6 +150,15 @@ This behavior allows that autoprovisioned parameters can freely established modi
 creation using the provisioning API. However, note that if a device (autoprovisioned or not) doesn't have these
 parameters defined at device level in database, the parameters are inherit from config group parameters.
 
+## Special measures and attributes names
+
+In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and 
+`measure_type`. The reason behind this is to avoid to collide with the original entity id and type, as mechanism
```
fgalan, line 156:
```suggestion
In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and 
`measure_type` attributes at Context Broker update. The reason behind this is to avoid to collide with the original entity id and type, as mechanism 
```

doc/api.md
mapedraza: The test that proves that statement is this one:
https://github.com/telefonicaid/iotagent-node-lib/blob/5073e6ae1e6afe608925213b7f624c26e9248edf/test/unit/ngsiv2/ngsiService/active-devices-test.js#L959-L1005

Being the provision, as you mentioned before:

```json
 { 
     name: 'type', 
     object_id: 't', 
     type: 'text' 
 }, 
 { 
     name: 'id', 
     object_id: 'i', 
     type: 'text' 
 }, 
```

The measure received (ignore the types, since it is because of using the IoTA Lib method to simulate the incoming measures):

```json
[ 
         { 
             name: 'i', 
             type: 'text', 
             value: 'idIoTA2' 
         }, 
         { 
             name: 't', 
             type: 'text', 
             value: 'typeIoTA2' 
         }, 
         { 
             name: 'm', 
             type: 'text', 
             value: 'measIoTA2' 
         } 
     ]; 
```

Being the mocked entity the following one:

```json
{ 
                 id: 'stupiddevice2', 
                 type: 'StupidDevice', 
                 meas: { 
                     value: 'measIoTA2', 
                     type: 'String' 
                 } 
             }
```

This means provisioned attributes named id or type are completely ignored.

fgalan: PR https://github.com/telefonicaid/iotagent-node-lib/pull/1568 probes @mapedraza 's point regarding provision of id or type is actually ignored.

Thus, this is NTC.

doc/api.md
```diff
@@ -149,6 +150,15 @@ This behavior allows that autoprovisioned parameters can freely established modi
...
 parameters defined at device level in database, the parameters are inherit from config group parameters.
 
+## Special measures and attributes names
+
+In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and 
+`measure_type` attributes at Context Broker update. The reason behind this is to avoid to collide with the original entity id and type, as mechanism 
+that enable store measure values from parameters called with the same name. It only applies to autoprovisioned
+attributes and is also available at JEXL context with the same name (`measure_id`or `measure_type`).
+
+In case of provisioning attributes using `id` or `type` as names, they are ignored.
```
fgalan, line 160:
```suggestion
In case of provisioning attributes using `id` or `type` as names (please don't do that ;), they are ignored. 
```

fgalan (approved):
LGTM

doc/api.md
```diff
@@ -149,6 +150,15 @@ This behavior allows that autoprovisioned parameters can freely established modi
 creation using the provisioning API. However, note that if a device (autoprovisioned or not) doesn't have these
 parameters defined at device level in database, the parameters are inherit from config group parameters.
 
+## Special measures and attributes names
+
+In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and 
+`measure_type` attributes at Context Broker update. The reason behind this is to avoid to collide with the original entity id and type, as mechanism
```
fgalan, line 156:
```suggestion
`measure_type` attributes at Context Broker update. The reason behind this is to avoid to collide with the original entity ID and type, as mechanism 
```

Issue #1578, opened 2024-02-09
add new API to delete high number of devices


AlvaroVega:

AlvaroVega:
Current delete is for one device:

i.e.: 
DELETE
	https://localhost/iota/iot/devices/DeviceID?apikey=gu1y31lo1agyr2xqf448n0s5v&protocol=IoTA-JSON

AlvaroVega:
@fgalan

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-node-lib/pull/1579

PR #1579, opened 2024-02-12
add multiple delete of devices

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-node-lib/issues/1578

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,2 @@
+- ADD: HTTP method to delete multiple devices at once (#1578)
 - Fix: store device subscriptions updates (#1086)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,2 +1,2 @@
-- ADD: HTTP method to delete multiple devices at once (#1578)
+- Add: `POST /iot/op/delete` operation to delete multiple devices at once (#1578)
 - Fix: store device subscriptions updates (#1086)
--- a/doc/api.md
+++ b/doc/api.md
@@ -1765,7 +1765,7 @@ Example:
 _**Response code**_
 
 -   `204` `OK` if successful.
--   `404` `NOT FOUND` if one or several devices were not found in the database.
+-   `404` `NOT FOUND` if one or several devices were not found in the database. Note that although a 404 error is returned the existing devices are deleted.
 -   `500` `SERVER ERROR` if there was any error not contemplated above.
 
 ## Miscellaneous API
--- a/doc/api.md
+++ b/doc/api.md
@@ -54,6 +54,7 @@
             -   [Get device details `GET /iot/devices/:deviceId`](#get-device-details-get-iotdevicesdeviceid)
             -   [Modify device `PUT /iot/devices/:deviceId`](#modify-device-put-iotdevicesdeviceid)
             -   [Remove device `DELETE /iot/devices/:deviceId`](#remove-device-delete-iotdevicesdeviceid)
+            -   [Remove devices `DELETE /iot/devices`](#remove-devices-delete-iotdevicesdeviceid)
     -   [Miscellaneous API](#miscellaneous-api)
         -   [Log operations](#log-operations)
             -   [Modify Loglevel `PUT /admin/log`](#modify-loglevel-put-adminlog)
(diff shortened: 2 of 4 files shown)


fgalan:
> A delete method with body is not fully restful but is not so ilegal, but we can look for other methods. @fgalan what do you think?

It is not illegal, but not sure if some network middleware (proxies, etc.) could have issues with this. Maybe it's better to avoid this risk using another approach.

Maybe we could use `POST /iot/op/delete`. This makes explicit that we are not using a REST operation (as we are not acting on a given resource, i.e. a device in IOAT API terms) but a kind of RPC operation (this is similar to the approach used in Orion API with the `POST /v2/op` operations family).

AlvaroVega:
> > A delete method with body is not fully restful but is not so ilegal, but we can look for other methods. @fgalan what do you think?
> 
> It is not illegal, but not sure if some network middleware (proxies, etc.) could have issues with this. Maybe it's better to avoid this risk using another approach.
> 
> Maybe we could use `POST /iot/op/delete`. This makes explicit that we are not using a REST operation (as we are not acting on a given resource, i.e. a device in IOAT API terms) but a kind of RPC operation (this is similar to the approach used in Orion API with the `POST /v2/op` operations family).

Taking a look to resource already used by device API:

https://github.com/telefonicaid/iotagent-node-lib/blob/7320146646757b6f39bdd0db052d1ca670e4f295/lib/services/northBound/deviceProvisioningServer.js#L438-L468

I was looking for something like POST `/iot/devices/delete`

CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
-- Fix: store device subscriptions updates (#1086)
\ No newline at end of file
+- ADD: HTTP method to delete multiple devices at once (#1578)
```
```suggestion
- Add: `POST /iot/op/delete` operation to delete multiple devices at once (#1578)
```

test/unit/ngsiv2/provisioning/removeProvisionedDevice-test.js
```diff
@@ -231,7 +231,42 @@ describe('NGSI-v2 - Device provisioning API: Remove provisioned devices', functi
...
+        it('should return a 204 OK and no errors', function (done) {
+            request(options, function (error, response, body) {
+                should.not.exist(error);
+                response.statusCode.should.equal(204);
+                done();
+            });
+        });
+    });
+
+    describe('When a request to remove a provision devices arrives', function () {
```
fgalan, line 243: Should functional test (i.e. the ones within /test/functional/) to be included?

doc/api.md
```diff
@@ -1720,10 +1723,51 @@ _**Request headers**_
...
+            "apikey": "apikey2"
+        }
+    ]
+}
+```
+
+_**Response code**_
+
+-   `204` `OK` if successful.
+-   `404` `NOT FOUND` if one or several devices were not found in the database.
```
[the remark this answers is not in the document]
AlvaroVega: IotAgent tries to delete all possible devices

test/unit/ngsiv2/provisioning/removeProvisionedDevice-test.js
AlvaroVega: Not sure if there is support for northbound api in functiona tests. @mapedraza is there support for that ?

doc/api.md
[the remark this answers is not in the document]
fgalan:
```suggestion
-   `404` `NOT FOUND` if one or several devices were not found in the database. Note that although a 404 error is returned the existing devices are deleted.
```

doc/api.md
```diff
@@ -1720,10 +1723,51 @@ _**Request headers**_
...
+-   `500` `SERVER ERROR` if there was any error not contemplated above.
```
fgalan, line 1769:
```suggestion
-   `500` `SERVER ERROR` if there was any error not considered above.
```

doc/api.md
```diff
@@ -152,12 +154,13 @@ parameters defined at device level in database, the parameters are inherit from
...
 ## Special measures and attributes names
 
-In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and 
-`measure_type` attributes at Context Broker update. The reason behind this is to avoid to collide with the original entity ID and type, as mechanism 
-that enable store measure values from parameters called with the same name. It only applies to autoprovisioned
-attributes and is also available at JEXL context with the same name (`measure_id`or `measure_type`).
+In case of arriving measures with name `id` or `type`, they are automatically transformed to `measure_id` and
+`measure_type` attributes at Context Broker update. The reason behind this is to avoid to collide with the original
+entity ID and type, as mechanism that enable store measure values from parameters called with the same name. It only
+applies to autoprovisioned attributes and is also available at JEXL context with the same name (`measure_id`or
```
fgalan, line 160:
```suggestion
applies to autoprovisioned attributes and is also available at JEXL context with the same name (`measure_id` or
```

test/unit/ngsiv2/provisioning/removeProvisionedDevice-test.js
mapedraza: So far, it is not supported northbound only tests (the tests consits of 3 steps, group provision, measure send and CB validation). No so hard to add support adding another "type" of test that check only the northbound using the same test infrastructure. I suggest to merge this like this and create an issue to implement what is describe in this comment.

fgalan: Issue created: https://github.com/telefonicaid/iotagent-node-lib/issues/1581 (please feel free to comment on it if something is missing/wrong)

NTC

fgalan (approved):
LGTM

PR #1591, opened 2024-03-26
fix(express):vulnerability

cecilia-par:


package.json
```diff
@@ -45,7 +45,7 @@
   "dependencies": {
     "async": "2.6.4",
     "body-parser": "~1.20.0",
-    "express": "~4.18.1",
+    "express": "^4.19.2",
```
fgalan, line 48:
```suggestion
    "express": "~4.19.2",
```

Our style is use ~ instead of ^ given that it leads to a more stable dependency.

package-lock.json
fgalan, line 1: We don't include packages-lock.json in our repositories. Please remove this file from PR.

package.json
fgalan: In addition, please add the following line in the CHANGES_NEXT_RELEASE file:

```
- Upgrade express dep from 4.18.1 to 4.19.2
```

fgalan (approved):
LGTM

Thanks for your contribution!

Issue #1557, opened 2024-01-24
Timeinstant as attribute case (not measure)


AlvaroVega:
### Is your feature request related to a problem? Please describe

When timestamp conf if false but:

attribute map is:
Param | Att Entity | type
Datetime | Timestamp | DateTime
TimeInstant | DateObserved | Datetime

measure contains: 
DateTIme:
TimeInstant: 

metadatas are lose

Full case is described at https://github.com/telefonicasc/jcyl-project/issues/153

AlvaroVega:
Testing with device with timestamp true and the following mapping

![Screenshot from 2024-02-23 11-15-02](https://github.com/telefonicaid/iotagent-node-lib/assets/983422/272ebf05-5eaf-46db-ab28-9ceb8120c664)

and with the follow measure:

 curl -i -X POST 'http://localhost:7897/iot/json?i=disp1&k=zrvj173zhvzb12t233z48vg96' -d '{ "TimeInstant": "2024-02-22T08:02:23.000Z", "datetime": "2024-02-22T08:02:23.000Z", "color": "blue"}' -H 'content-type: application/json'


entity has provided TimeInstant in metadata attributes:

![Screenshot from 2024-02-23 11-16-19](https://github.com/telefonicaid/iotagent-node-lib/assets/983422/362c8723-6908-46bb-8c9c-4d3b46977561)

AlvaroVega:
Testing with device with timestamp false and the following mapping

![Screenshot from 2024-02-23 11-15-02](https://github.com/telefonicaid/iotagent-node-lib/assets/983422/5eeaab8f-b0a7-4ec2-9ee5-bbb752f69ab1)


and with the follow measure:

curl -i -X POST 'http://localhost:7897/iot/json?i=disp1&k=zrvj173zhvzb12t233z48vg96' -d '{ "TimeInstant": "2024-02-22T08:02:23.000Z", "datetime": "2024-02-22T08:02:23.000Z", "color": "blue"}' -H 'content-type: application/json'

entity has provided TimeInstant in metadata attributes:

![Screenshot from 2024-02-23 11-18-10](https://github.com/telefonicaid/iotagent-node-lib/assets/983422/09bf3b48-6675-476d-8223-3cc66fe54b45)

AlvaroVega:
Testing with a group with timeinstant=true and the following measures:

[...]

PR #1593, opened 2024-04-03
add test case about attribute mapped to timestamp

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/issues/1557

### DIFF
--- a/test/functional/testCases.js
+++ b/test/functional/testCases.js
@@ -1036,7 +1036,8 @@ const testCases = [
         ]
     },
     {
-        describeName: '0140 - Simple group with active attribute + chained JEXL expression text (a | trim | replacestr("hello","hi"))',
+        describeName:
+            '0140 - Simple group with active attribute + chained JEXL expression text (a | trim | replacestr("hello","hi"))',
         provision: {
             url: 'http://localhost:' + config.iota.server.port + '/iot/services',
             method: 'POST',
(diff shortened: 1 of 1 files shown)


AlvaroVega:
We should include tests about include also:
- explicitAttrs derivations of these cases
- timestamp measure
And maybe move block where timestamp is processed to the end

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -552,6 +505,31 @@ function sendUpdateValueNgsi2(entityName, measures, typeInformation, token, call
...
+                    timestamp.value = timestampAttrs[0]['value'];
+                }
+
+                if (timestamp.value) {
+                    if (!moment(timestamp.value, moment.ISO_8601, true).isValid()) {
+                        callback(new errors.BadTimestamp(timestamp.value, entityName, typeInformation));
+                        return;
+                    }
+                } else {
+                    if (!typeInformation.timezone) {
```
mrutid, line 524: Maybe this could be precalculated to simplify logic and optimize. Minor.

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -552,6 +505,31 @@ function sendUpdateValueNgsi2(entityName, measures, typeInformation, token, call
...
+                        timestamp.value = new Date().toISOString();
+                        jexlctxt[constants.TIMESTAMP_ATTRIBUTE] = timestamp.value; // nosense
+                    } else {
+                        timestamp.value = moment().tz(typeInformation.timezone).format('YYYY-MM-DD[T]HH:mm:ss.SSSZ');
+                        jexlctxt[constants.TIMESTAMP_ATTRIBUTE] = timestamp.value; // nosense
```
AlvaroVega, line 529: jexlctxt is not used any more from this point

lib/services/ngsi/entities-NGSI-v2.js
AlvaroVega: Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1593/commits/4c8a62478a645ff66218e73012b3b52c0c9da833

doc/api.md
```diff
@@ -972,32 +972,37 @@ Will now generate the following NGSI v2 payload:
...
+The `timestamp` conf value used is:
 
 -   The one defined at device level
 -   The one defined at group level (if not defined at device level)
 -   The one defined at [IoTA configuration level](admin.md#timestamp) / `IOTA_TIMESTAMP` env var (if not defined at
     group level or device level)
 
+Some additional considerations to take into account:
+
+* If there is an attribute which maps a measure to `TimeInstant` attribute (after [expression evaluation](#expression-language-support) if any is defined), then that value will be used as as `TimeInstant,
```
fgalan, line 1002:
```suggestion
* If there is an attribute which maps a measure to `TimeInstant` attribute (after [expression evaluation](#expression-language-support) if any is defined), then that value will be used as `TimeInstant,
```

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,5 @@
+- Fix: TimeInstant mapped from measure overrides system TimeInstant (#1557)
```
```suggestion
- Fix: TimeInstant mapped from attribute overrides default behaviours (#1557)
```

fgalan (approved):
LGTM

PR #1605, opened 2024-04-30
Add tests using explicitAttrs with value `[ ]`

mapedraza:
Adding tests described in #1600 

- 0531 - Group with explicit attrs:[ ] (empty array) + active attributes + static attributes
   - A - WHEN sending both provisioned and not object_ids (measures) through http IT should NOT store anything into the Context Broker (No request to CB)
- 0532 - Group with explicit attrs:[ ] (empty array) + active attributes + static attributes + timestamp:true
   - A - WHEN sending both provisioned and not object_ids (measures) through http IT should NOT store anything into the Context Broker (No request to CB)
   - B - WHEN sending data and a measure called TimeInstant through http IT should NOT store anything into the Context Broker (No request to CB)
- 0533 - Group with explicit attrs:[ ] (empty array) + active attributes + TimeInstant attribute + static attributes + timestamp:true
   - A - WHEN sending data and a measure called "t" (defined as TimeInstant attribte) through http IT should NOT store anything into the Context Broker (No request to CB)
   - B - WHEN sending data and a measure called TimeInstant through http IT should NOT store anything into the Context Broker (No request to CB)

BUT some problems were discovered when using TimeInstant attribute and fucntionalities in cases 533 A) and B):

**0533 B**
``` diff
Assertion Error - Context Broker received payload differs from expectation
+ expected - actual

-{
-  "TimeInstant": {
-    "type": "DateTime"
-    "value": "2015-12-14T08:06:01.468Z"
-  }
-  "id": "TestType:TestDevice"
-  "type": "TestType"
-}
+{}
```

**0533 B**
``` diff
Assertion Error - Context Broker received payload differs from expectation
+ expected - actual

-{
-  "DateIssued": {
-    "metadata": {
-      "TimeInstant": {
-        "type": "DateTime"
-        "value": "2024-04-29T16:07:27.281Z"
-      }
-    }
-    "type": "DateTime"
-    "value": "2015-12-14T08:06:01.468Z"
-  }
-  "TimeInstant": {
-    "type": "DateTime"
-    "value": "2024-04-29T16:07:27.281Z"
-  }
-  "id": "TestType:TestDevice"
-  "type": "TestType"
-}
+{}
```

Issue #1606, opened 2024-04-30
Problems with explicitAttrs and TimeInstan #1605


mapedraza:
### Unexpected behaviour you saw

When using explictAttrs with value `[]` (no attribute should pass), TimeInstant value pass. See https://github.com/telefonicaid/iotagent-node-lib/pull/1605

AlvaroVega:
corner case?

fgalan:
PR #1610 is the regression for this issue (i.e. when this issue gets fixed, the test included in that PR will check it). In fact, maybe this issue could be solved in task/add-533-func branch (the one used by PR #1610) so we have all in the same place.

mapedraza:
> I understand that after merging PR #1618 this issue can be closed.

That's right

PR #1610, opened 2024-05-14
Add TimeInstant + explicitAttr tests (ftest 533)

mapedraza:
A part of original PR https://github.com/telefonicaid/iotagent-node-lib/pull/1605 including test 533

>    0533 - Group with explicit attrs:[ ] (empty array) + active attributes + TimeInstant attribute + static 
>    - A - WHEN sending data and a measure called "t" (defined as TimeInstant attribte) through http IT should NOT 
>    - B - WHEN sending data and a measure called TimeInstant through http IT should NOT store anything into the 
> 
> 
> BUT some problems were discovered when using TimeInstant attribute and fucntionalities in cases 533 A) and B):
> 
> **0533 A**
> ``` diff
> Assertion Error - Context Broker received payload differs from expectation
> + expected - actual
> 
> -{
> -  "TimeInstant": {
> -    "type": "DateTime"
> -    "value": "2015-12-14T08:06:01.468Z"
> -  }
> -  "id": "TestType:TestDevice"
> -  "type": "TestType"
> -}
> +{}
> ```
> 
> **0533 B**
> ``` diff
> Assertion Error - Context Broker received payload differs from expectation
> + expected - actual
> 
> -{
> -  "DateIssued": {
> -    "metadata": {
> -      "TimeInstant": {
> -        "type": "DateTime"
> -        "value": "2024-04-29T16:07:27.281Z"
> -      }
> -    }
> -    "type": "DateTime"
> -    "value": "2015-12-14T08:06:01.468Z"
> -  }
> -  "TimeInstant": {
> -    "type": "DateTime"
> -    "value": "2024-04-29T16:07:27.281Z"
> -  }
> -  "id": "TestType:TestDevice"
> -  "type": "TestType"
> -}
> +{}
> ```

See https://github.com/telefonicaid/iotagent-node-lib/pull/1605#issuecomment-2100832653

### DIFF
--- a/test/functional/testCases.js
+++ b/test/functional/testCases.js
@@ -2970,6 +2970,107 @@ const testCases = [
     },
     {
         describeName:
+            '0533 - Group with explicit attrs:[ ] (empty array) + active attributes + TimeInstant attribute + static attributes + timestamp:true',
+        provision: {
+            url: 'http://localhost:' + config.iota.server.port + '/iot/services',
+            method: 'POST',
+            json: {
+                services: [
+                    {
+                        resource: '/iot/json',
+                        apikey: globalEnv.apikey,
+                        entity_type: globalEnv.entity_type,
+                        explicitAttrs: "['DateIssued']",
+                        timestamp: true,
+                        commands: [],
+                        lazy: [],
+                        attributes: [
+                            {
+                                name: 'attr_a',
+                                object_id: 'a',
+                                type: 'Number'
+                            },
+                            {
+                                name: 'attr_b',
+                                object_id: 'b',
+                                type: 'Number'
+                            },
+                            {
+                                name: 'DateIssued',
+                                object_id: 'TimeInstant',
+                                type: 'DateTime'
+                            },
+                            {
+                                name: 'TimeInstant',
+                                object_id: 't',
+                                type: 'DateTime'
+                            }
+                        ],
+                        static_attributes: [
+                            {
+                                name: 'static_a',
+                                type: 'Number',
+                                value: 3
+                            },
+                            {
+                                name: 'static_b',
+                                type: 'Number',
+                                value: 4
+                            }
+                        ]
+                    }
+                ]
+            },
+            headers: {
+                'fiware-service': globalEnv.service,
+                'fiware-servicepath': globalEnv.servicePath
+            }
+        },
+        should: [
+            {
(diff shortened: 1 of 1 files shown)


fgalan:
This is a future regression case for issue #1606

test/functional/testCases.js
```diff
@@ -2968,6 +2968,107 @@ const testCases = [
...
+        provision: {
+            url: 'http://localhost:' + config.iota.server.port + '/iot/services',
+            method: 'POST',
+            json: {
+                services: [
+                    {
+                        resource: '/iot/json',
+                        apikey: globalEnv.apikey,
+                        entity_type: globalEnv.entity_type,
+                        explicitAttrs: "['DateIssued']",
```
AlvaroVega, line 2983: @mapedraza this explicitAttrs is not empty as suggest description case, is it?

test/functional/testCases.js
```diff
@@ -2968,6 +2968,107 @@ const testCases = [
...
+                        k: globalEnv.apikey
+                    },
+                    json: {
+                        a: 3,
+                        b: 10,
+                        c: 11,
+                        TimeInstant: '2015-12-14T08:06:01.468Z'
+                    }
+                },
+                expectation: {} // No payload expected
```
AlvaroVega, line 3068: No CB update is represented in other testCases (like 0120, ) as null array:
```
expectation: []
```

AlvaroVega:
IMHO with this group definition:
```
                        explicitAttrs: "[ ]",
                        timestamp: true,
```
a TimeInstant will be always propaged to CB

AlvaroVega:
> IMHO with this group definition:
> 
> ```
>                         explicitAttrs: "[ ]",
>                         timestamp: true,
> ```
> 
> a TimeInstant will be always propaged to CB, regardless attribute mapping definition.
> 
> This PR is as I think it would be: #1614

If the logic is that `explicitAttrs` empty implies that no `timestamp` should be propagated to CB then the logic to change is

https://github.com/telefonicaid/iotagent-node-lib/blob/3e3842a14b77664c10aa896b31d157d267e36493/lib/services/ngsi/entities-NGSI-v2.js#L524

mapedraza:
overpassed by https://github.com/telefonicaid/iotagent-node-lib/pull/1618

PR #1607, opened 2024-05-06
allow to send to CB batch update for multimeasures (and sort them by TimeInstant)

AlvaroVega:
related to fix https://github.com/telefonicaid/iotagent-json/issues/825

- [x] Add functional test about a multimeasure
- [x] Add functional test about a multimeasure sorted by TimeInstant
- [x] Update Doc

PR #1623, opened 2024-06-25
Allow to send to CB batch update for multimeasures for NGSI-LD

jason-fox:
Equivalent of #1607 - it is just a copy of `sendUpdateValueNgsi2()` for now with relevant changes for NGSI-LD.

At some point the v2 functionality had been updated to : `Not update value in the Context Broker  with an empty payload`.
This path is more frequent in NGSI-LD as attributes which resolve to `null` are already removed. This leads to the update of a large number of test expectations where an "empty entity" was being appended to the upsert.

Several (mainly JEXL) tests are no longer relevant, since they result in a `null` and no CB update - I have currently set those tests to PENDING for now pending further review - I can delete them if necessary.


AlvaroVega (approved):
LGTM

fgalan:
An entry in CHANGES_NEXT_RELEASE describing the change in this PR (including issue reference, if any) would be great.

jason-fox:
> I think it would be better to delete them, for cleanness, please.

Fixed  079bb44

> An entry in CHANGES_NEXT_RELEASE

Fixed 8de86c7

fgalan (approved):
LGTM

Issue #1630, opened 2024-08-02
Include attribute metadata as part of the JEXL context


mapedraza:
### Component

IoT Agent Node Lib
IoT Agent JSON

### Is your feature request related to a problem? Please describe

Include attribute metadata as part of the JEXL context

Related implementation in CB: https://github.com/telefonicaid/fiware-orion/issues/4594#issuecomment-2264980414

### Describe the solution you'd like

Include attribute metadata as part of the JEXL context

File related:
https://github.com/telefonicaid/iotagent-node-lib/blob/46ac1d88a392f4432d4c849c2522c1d6a577f285/lib/services/ngsi/entities-NGSI-v2.js#L340

Note: In the case of NGSIv2 and NGSI-LD payloads, it would be great to have the payload metadata context available before evaluating attributes expression.

A uses case is to use `previousValue` built-in metadata to evaluate if an attribute really changes and, eventually, propagate or not the IoTA update depending on this (to avoid overload the destination CB)

fgalan:
> A uses case is to use `previousValue` built-in metadata to evaluate if an attribute really changes and, eventually, propagate or not the IoTA update depending on this (to avoid overload the destination CB)

That would be done by a subscription at CB with `"metadata": [ "previousValue"]" using IOTA southbound API as endpoint.

fgalan:
PR https://github.com/telefonicaid/iotagent-node-lib/pull/1631

PR #1631, opened 2024-08-05
Task/add attribute metatadas and static attribute metadatas to jexl ctxt

AlvaroVega:
Issue https://github.com/telefonicaid/iotagent-node-lib/issues/1630

Implemented having into account: 
https://github.com/telefonicaid/fiware-orion/pull/4600/

- [x] Add test about metadata in jexl context

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -596,7 +596,7 @@ Attribute metadata and Static Attribute metadata are available in the **context*
 `metadata.<AttributeName>.<MetadataName>` or `metadata.<StaticAttributeName>.<MetadataName>` in a similar way of defined
 for [Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
 
-Moreover if attribute metdata has an expression metadata attribute value in jexl context is updated after that
+Moreover, if attribute metadata has an expression metadata attribute value in jexl context it is updated after that
 expression is evaluated. Note that there is no order into metadata structure and there is no warranty about which
 metadata attribute expression will be evaluated first.
 
--- a/doc/api.md
+++ b/doc/api.md
@@ -593,8 +593,8 @@ Additionally, for attribute expressions (`expression`, `entity_name`), `entityNa
 (`expression`) measures are available in the **context** used to evaluate them.
 
 Attribute metadata and Static Attribute metadata are available in the **context** under the following convention:
-`metadata_AttributeName_MetadataName` or `metadata_StaticAttributeName_MetadataName` in a similar way of defined for
-[Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
+`metadata_<AttributeName>_<MetadataName>` or `metadata_<StaticAttributeName>_<MetadataName>` in a similar way of defined
+for [Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
 
 ### Examples of JEXL expressions
 
--- a/doc/api.md
+++ b/doc/api.md
@@ -593,7 +593,7 @@ Additionally, for attribute expressions (`expression`, `entity_name`), `entityNa
 (`expression`) measures are available in the **context** used to evaluate them.
 
 Attribute metadata and Static Attribute metadata are available in the **context** under the following convention:
-`metadata.AttributeName.MetadataName` or `metadata.StaticAttributeName.MetadataName` in a similar way of defined for
+`metadata_AttributeName_MetadataName` or `metadata_StaticAttributeName_MetadataName` in a similar way of defined for
 [Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
 
 ### Examples of JEXL expressions
--- a/doc/api.md
+++ b/doc/api.md
@@ -245,10 +245,13 @@ measure name, unless `explicitAttrs` is defined. Measures `id` or `type` names a
 ## Device autoprovision and entity creation
 
 For those agents that uses IoTA Node LIB version 3.4.0 or higher, you should consider that the entity is not created
-automatically when a device is created. This means that all entities into the Context Broker are created when data 
-arrives from a device, no matter if the device is explicitly provisioned (via [device provisioning API](#create-device-post-iotdevices)) or autoprovisioned.
+automatically when a device is created. This means that all entities into the Context Broker are created when data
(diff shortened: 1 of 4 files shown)


mapedraza:
Should context also need to be updated while calculating metadata expressions here?
https://github.com/telefonicaid/iotagent-node-lib/blob/49f85502765ed7b801de68233aef4fe700820b85/lib/services/ngsi/entities-NGSI-v2.js#L506-L537

mapedraza:
Would be great to add a test under JEXL section here:
https://github.com/telefonicaid/iotagent-node-lib/blob/46ac1d88a392f4432d4c849c2522c1d6a577f285/test/functional/testCases.js#L987

fgalan:
> Should context also need to be updated while calculating metadata expressions here?
> 
> https://github.com/telefonicaid/iotagent-node-lib/blob/49f85502765ed7b801de68233aef4fe700820b85/lib/services/ngsi/entities-NGSI-v2.js#L506-L537

Differently from attribute calculation (where there is an order based in array index) in the case of metadata we don't have such array, so we have to be careful if the calculation of one metadata depends on another calculated metadata in the same attribute. Maybe this such be warned at documentation.

doc/api.md
```diff
@@ -588,6 +592,10 @@ expression. In all cases the following data is available to all expressions:
 Additionally, for attribute expressions (`expression`, `entity_name`), `entityNameExp` and metadata expressions
 (`expression`) measures are available in the **context** used to evaluate them.
 
+Attribute metadata and Static Attribute metadata are available in the **context** under the following convention:
+`metadata.AttributeName.MetadataName` or `metadata.StaticAttributeName.MetadataName` in a similar way of defined for
+[Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
```
fgalan, line 597: It seems this paragraph is the only new thing in documentation, all the rest seems to be re-formatting.

@AlvaroVega could you confirm, pls?

fgalan:
```suggestion
Attribute metadata and Static Attribute metadata are available in the **context** under the following convention:
`metadata.<AttributeName>.<MetadataName>` or `metadata.<StaticAttributeName>.<MetadataName>` in a similar way of defined for
[Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
```

Using the `<...>` syntax would make more explicit that they refer to actual names, different to `metadata` which is a literal token.

AlvaroVega: Yes, it is

AlvaroVega:
I've to update format for jexl context  from 'metadata.<AttributeName>.<MetadataName>' to `metadata_<AttributeName>_<MetadataName>` due to '.' imply array access if is used in a jexl expression

AlvaroVega:
> Would be great to add a test under JEXL section here:
> 
> https://github.com/telefonicaid/iotagent-node-lib/blob/46ac1d88a392f4432d4c849c2522c1d6a577f285/test/functional/testCases.js#L987-L997

Fixed in https://github.com/telefonicaid/iotagent-node-lib/pull/1631/commits/c8778c79c495b3c4236a7fbc5f7443eb86638f51

AlvaroVega:
> Should context also need to be updated while calculating metadata expressions here?
> 
> https://github.com/telefonicaid/iotagent-node-lib/blob/49f85502765ed7b801de68233aef4fe700820b85/lib/services/ngsi/entities-NGSI-v2.js#L506-L537

Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1631/commits/344cb5c66ec74d84ae8270bd70d77b18d61c4bd6

AlvaroVega:
> Differently from attribute calculation (where there is an order based in array index) in the case of metadata we don't have such array, so we have to be careful if the calculation of one metadata depends on another calculated metadata in the same attribute. Maybe this such be warned at documentation.

Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1631/commits/a103f39312bc1bad5f48dbc1a9fdafecba15852e

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -511,6 +535,11 @@ function sendUpdateValueNgsi2(entityName, originMeasures, originTypeInformation,
                         }
                         newAttrMeta['value'] = metaValueExpression;
                         currentAttr.metadata[metaKey] = newAttrMeta;
+
+                        //RE-Populate de JEXLcontext
+                        if (jexlctxt.metadata && jexlctxt.metadata[currentAttr.name]) {
```
mapedraza, line 540: This `if` clause checks if the metadata exits into the original jexl context but, in the case of new metadatas generated as a result of expressions, it would not be included. I think It should be appended to the jexlctxt even if it was not part previously of it.

mapedraza (approved):
LGTM

doc/api.md
```diff
@@ -588,6 +592,14 @@ expression. In all cases the following data is available to all expressions:
 Additionally, for attribute expressions (`expression`, `entity_name`), `entityNameExp` and metadata expressions
 (`expression`) measures are available in the **context** used to evaluate them.
 
+Attribute metadata and Static Attribute metadata are available in the **context** under the following convention:
+`metadata.<AttributeName>.<MetadataName>` or `metadata.<StaticAttributeName>.<MetadataName>` in a similar way of defined
+for [Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
+
+Moreover if attribute metdata has an expression metadata attribute value in jexl context is updated after that
+expression is evaluated. Note that there is no order into metadata structure and there is no warranty about which
+metadata attribute expression will be evaluated first.
```
fgalan, line 601:
```suggestion
Moreover, if attribute metadata has an expression metadata attribute value in jexl context it is updated after that
expression is evaluated. Note that there is no order into metadata structure and there is no warranty about which
metadata attribute expression will be evaluated first.
```

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -273,6 +273,26 @@ function sendUpdateValueNgsi2(entityName, originMeasures, originTypeInformation,
             return initObj;
         }
     }
+    //it returns a flat object from an Metadata Object
```
fgalan, line 276: Is this comment still valid? What "flat" refers here?

Maybe better this one?

```suggestion
    //it returns a metadata object using the same structure described
    // at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support
```

fgalan (approved):
LGTM

PR #1523, opened 2023-11-09
[Bug?] Using device data JEXL context when executing `iotAgentLib.update` into tests

mapedraza:
When executing tests in IoTA Lib as usual way ( using `iotAgentLib.update` method ), the device JEXL context does not include parameters such as service, subservice or device ID. 

I do not know if this is something that need to be fixed, but, at least, it should be documented. Found when migrating https://github.com/telefonicaid/iotagent-json/pull/769 to this repo

PR #1636, opened 2024-08-07
access to id, service and subservice in jexl context for functional test

AlvaroVega:
fix for https://github.com/telefonicaid/iotagent-node-lib/pull/1523

### DIFF
--- a/test/functional/testCases.js
+++ b/test/functional/testCases.js
@@ -1784,7 +1784,7 @@ const testCases = [
                                 object_id: 'a',
                                 name: 'attr_a',
                                 type: 'Text',
-                                expression: 'a+":"+service+subservice+type'
+                                expression: 'a+":"+service+subservice+id+type'
                             }
                         ]
                     }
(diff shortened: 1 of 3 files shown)


test/functional/testCases.js
```diff
@@ -1785,7 +1784,7 @@ const testCases = [
                                 object_id: 'a',
                                 name: 'attr_a',
                                 type: 'Text',
-                                expression: 'a+":"+service+subservice+id+type'
```
AlvaroVega, line 1788: groupToIoTAConfigType is not providen deviceId

Maybe  testUtils.testCase should do that

fgalan:
I think it would be wise to use same literals than the one used at CB (https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#macro-substitution)

> * `id`: for the `id` of the entity
> * `type`: for the `type` of the entity
> * `service`: for the service (i.e. `fiware-service` header value) in the
>   update request triggering the subscription.
> * `servicePath`: for the service path (i.e. `fiware-servicepath` header value) in the
>   update request triggering the subscription.
> * `authToken: for the authorization token (i.e. `x-auth-token` header value) in the
>   update request triggering the subscription.
> * All the attributes in the entity triggering the notification (included in the update triggering the notification or not)

Is the PR working that way?

AlvaroVega:
> Is the PR working that way?

Not. This is just to align functional tests with implemented feature in iotagent-node-lib, specially test about jexl context

fgalan:
> > Is the PR working that way?
> 
> Not. This is just to align functional tests with implemented feature in iotagent-node-lib, specially test about jexl context

CB uses `servicePath`. IOTA uses `subservice`. They are different but probably is not a good idea to change any of them to align, due to potential backward compatibility problems and changes in the IOTA DB model.

mapedraza (approved):
LGTM

Issue #1634, opened 2024-08-06
[DOC] Add to special measure name `metadata` key


mapedraza:
https://github.com/telefonicaid/iotagent-node-lib/blob/c57d8929d4a44c7ee57ba16d673024b8ea3ffb75/doc/api.md?plain=1#L169-L177

EDIT:

Add also `metadata` here:
https://github.com/telefonicaid/iotagent-node-lib/blob/c57d8929d4a44c7ee57ba16d673024b8ea3ffb75/doc/api.md#L593-L602

AlvaroVega:
done in https://github.com/telefonicaid/iotagent-node-lib/pull/1641. Should we close this?

PR #1641, opened 2024-08-08
FIX metadata expression evaluation in doc

fgalan:
Maybe related (not sure) #1634

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -30,6 +30,7 @@
     -   [Measurement transformation](#measurement-transformation)
         -   [Measurement transformation definition](#measurement-transformation-definition)
         -   [Measurement transformation execution](#measurement-transformation-execution)
+        -   [Metadata transformation](#metadata-transformation)
         -   [Measurement transformation order](#measurement-transformation-order)
         -   [Multientity measurement transformation support (`object_id`)](#multientity-measurement-transformation-support-object_id)
     -   [Command execution](#command-execution)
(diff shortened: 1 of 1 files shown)


doc/api.md
```diff
@@ -607,16 +569,13 @@ expression. In all cases the following data is available to all expressions:
...
 
-Attribute metadata and Static Attribute metadata are available in the **context** under the following convention:
+- measures, as `<AttributeName>`
+- metadata (both for attribute and static attribute) are available in the **context** under the following convention:
 `metadata.<AttributeName>.<MetadataName>` or `metadata.<StaticAttributeName>.<MetadataName>` in a similar way of defined
 for [Context Broker](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#metadata-support)
 
-Moreover, if attribute metadata has an expression metadata attribute value in jexl context it is updated after that
-expression is evaluated. Note that there is no order into metadata structure and there is no warranty about which
-metadata attribute expression will be evaluated first.
```
fgalan, line 618: > Moreover, if attribute metadata has an expression metadata attribute value in jexl context it is updated after that
expression is evaluated

I think this is redundant with the contents of "Measurement tranformation" section, so it is removed.

> Note that there is no order into metadata structure and there is no warranty about which
metadata attribute expression will be evaluated first.

This is about evaluation ordering, so moved to "Measurement transformation order" section.

doc/api.md
```diff
@@ -607,16 +570,13 @@ expression. In all cases the following data is available to all expressions:
 -   `staticAttributes`: static attributes defined in the device or config group
 
 Additionally, for attribute expressions (`expression`, `entity_name`), `entityNameExp` and metadata expressions
-(`expression`) measures are available in the **context** used to evaluate them.
+(`expression`) the following is available in the **context** used to evalute:
 
-Attribute metadata and Static Attribute metadata are available in the **context** under the following convention:
+- measures, as `<AttributeName>`
+- metadata (both for attribute measurement in the case of NGSIv2 measurments and static attribute) are available in the **context** under the following convention:
```
fgalan, line 576:
```suggestion
- metadata (both for attribute measurement in the case of NGSI-v2 measurements and static attribute) are available in the **context** under the following convention:
```

AlvaroVega (approved):
LGTM

mapedraza (approved):
LGTM

Issue #1633, opened 2024-08-06
[DOC] Uniqueness of groups and devices


mapedraza:
Is this correct?

https://github.com/telefonicaid/iotagent-node-lib/blob/c57d8929d4a44c7ee57ba16d673024b8ea3ffb75/doc/api.md?plain=1#L162-L167

Is not group defined by apikey and resource?
Devices by device_id and apikey?

CC: @fgalan @AlvaroVega

AlvaroVega:
According with model/Group.js and lib/services/groups/*  `Group service uniqueness is defined by the combination of: service, subservice, resource and apikey `.
I don't why but I've always use default `resource`, but it`s possible to use different resources.
Device uniqueness is as described: service, subservice, device_id and apikey.

fgalan:
Should this issue be closed after the merge of PR https://github.com/telefonicaid/iotagent-node-lib/pull/1639 ?

mapedraza:
Yes, it is solved after #1639

PR #1592, opened 2024-04-03
Task/allow devices same device id in same path but apikey

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/issues/1589

Related with https://github.com/telefonicaid/iotagent-node-lib/issues/1426 https://github.com/telefonicaid/iotagent-node-lib/issues/1435) (https://github.com/telefonicaid/iotagent-node-lib/issues/1430)

Issue #1589, opened 2024-03-13
Wrong association of measurements to the correct device using different service groups but same device_id


tudamp:
### IoT Agent Node Lib version the issue has been seen with

4.1.0

### Bound or port used (API interaction)

Southbound (Devices data API)

### NGSI version

NGSIv2

### Are you running a container?

Yes, I am using a contaner (Docker, Kubernetes...)

### Image type

distroless

### Expected behaviour you didn't see

By creating an environment using service groups with these settings:
| **Parameters**         | **Type1:Sensor01** | **Type2:Sensor01** |   
|--------------------|----------------|----------------|
| Fiware-Service     | service1       | service1       |   
| Fiware-ServicePath | /              | /              |   
| Resource           | /iot/json      | /iot/json      |   
| API Key            | apikey01       | apikey02       |  
| Entity Type        | Type1          | Type2          |   
| Device ID          | Sensor01         | Sensor01         |   
| Entity Name        | Type1:Sensor01 | Type2:Sensor01 | 

By sending the measurements first of Type1:Sensor01 with apikey1 and then of Type2:Sensor01 with its apikey2, the agent should:

1. recognise the service group of the first send (using apikey01 and resource), create the device if not present with the correct Entity type of the service group, send the measurement to cb

2.  repeat the same behaviour with the second data send from the  Type2:Sensor01

### Unexpected behaviour you saw

The step1:
1. recognise the service group of the first send (using apikey01 and resource), create the device if not present with the correct Entity type of the service group, send the measurement to cb

occurs correctly: the service group matched to the apikey is recognised and the settings (type1) combined with the measurement arriving from the device.

In step 2, the service group matched to apikey2 is recognised, but given the presence of a device with the same id, the measurement is combined with this device, resulting in an incorrect match with the entity in the cb (type1 instead of type2).
The device in the second service group is never created.

### Configs

```yaml
environment:
    - "IOTA_CB_HOST=orion"
    - "IOTA_CB_PORT=1026"
    - "IOTA_NORTH_PORT=4041"
    - "IOTA_REGISTRY_TYPE=mongodb"
    - "IOTA_MONGO_HOST=mongodb"
    - "IOTA_MONGO_PORT=27017"
    - "IOTA_MONGO_DB=iotagent-json"
    - "IOTA_HTTP_PORT=7896"
    - "IOTA_CB_NGSI_VERSION=v2"
    - "IOTA_DEFAULT_RESOURCE=/iot/json"
[...]

PR #1666, opened 2024-10-21
fix doc uniqueness group

AlvaroVega:
related with #1633 #1589

Given a measure (identified by an apikey) there is no way to identify group related if apikey is not unique for all services and subservices.

My comment https://github.com/telefonicaid/iotagent-node-lib/issues/1633#issuecomment-2272722185 was  taken directly from https://github.com/telefonicaid/iotagent-node-lib/pull/1592#discussion_r1551377609


fgalan (approved):
LGTM

Issue #1660, opened 2024-09-30
Some data are not reported to Orion when updating a device


GuillaumeOd:
### IoT Agent Node Lib version the issue has been seen with

4.6.0

### Bound or port used (API interaction)

Southbound (Devices data API)

### NGSI version

NGSIv2

### Are you running a container?

Yes, I am using a contaner (Docker, Kubernetes...)

### Image type

normal

### Expected behaviour you didn't see

When upgrading iot-agent-node-lib from 3.3.4 to 4.6.0

It should correctly updating a device attribute when using `iotAgentLib.update(device.name, device.type, "", [ { name: 'value', type: 'number', value: '10' }], device)`

iot-agent-node-lib 3.3.4 is still working perfectly

### Unexpected behaviour you saw

The payload request made to Orion is perfectly sent but does not contain de requested change on the device attribute

Here's what is sent to Orion in 3.3.4 :

```sh
{ op: 'IoTAgentNGSI.Request' } Options: %s {
    "url": "http://orion:1026/v2/entities?options=upsert",
    "method": "POST",
    "headers": {
        "fiware-service": "ngeconnect",
        "fiware-servicepath": "/"
    },
    "json": {
        "id": "urn:ngsild:TemperatureSensor:CypressDeviceA1A",
        "type": "TemperatureSensor",
        "value": {
            "value": "10",
            "type": "number",
            "metadata": {
                "TimeInstant": {
                    "type": "DateTime",
                    "value": "2024-09-30T08:41:12.495Z"
                }
            }
        }
        (.........)
    }
}
```

In 4.6.0 the payload is exactly the same but the "value" attribute in the request has disappeared, so the request is sent to Orion but without the updated value on the device. End to end, this leads to having the exact same device in orion as the upsert request did not sent the attribute to update.

I believe this comes from `sendUpdateValueNgsi2()` in `entities-NGSI-v2.js` as this function changed a lot between 3.3.4 and 4.6.0. 

After some debugging and analysis, it's coming from the line `measures = measures.filter((item) => item.name !== currentAttr.object_id && item.name !== currentAttr.name);` that removes the measures `value`. The second part of the filter `item.name !== currentAttr.name` is going to filter all attributes names that are present in `preprocessedAttr` but it contains `typeInformation.active` from `originTypeInformation` which contains the `value` attribute...

### Configs

```yaml
environment:
    - "IOTA_CB_HOST=orion"
    - "IOTA_CB_PORT=1026"
[...]

PR #1667, opened 2024-11-05
fix: attribute loss when updating entity

GuillaumeOd:
related to https://github.com/telefonicaid/iotagent-node-lib/issues/1660


AlvaroVega (dismissed):
As a commented here https://github.com/telefonicaid/iotagent-node-lib/issues/1660#issuecomment-2457001646 a test case about his issue is needed.

fgalan:
@GuillaumeOd please have a look to documentation in the case it has to change to describe the new behaviour with this PR.

AlvaroVega:
This PR is reverting this commit: https://github.com/telefonicaid/iotagent-node-lib/commit/166f7715ee612c1a583e30d2d44d123ee90f261c

fgalan:
@GuillaumeOd not sure if I'm getting the point...

You want to send measures like this:

```
{
  "fireDetectionThreshold": 10
}
```

and get corresponding attribute at CB this way:

```
                    "fireDetectionThreshold": {
                        type: "Number",
                        value: 10
                    }
```

is my interpretation correct?

GuillaumeOd:
@fgalan yes exactly, for the type I'm not sure if it should be `Number `or `Text` by the way, but your interpretation  is correct

fgalan:
> @fgalan Yes, exactly! Regarding the type, I'm not entirely sure if it should be `Number` or `Text`, but your interpretation seems spot on.

In that case, why don't use at provisioning time

```
                            {
                                name: 'fireDetectionThreshold',
                                type: 'Number'
                            }
```

instead of

```
                            {
                                object_id: 'f_dt',
                                name: 'fireDetectionThreshold',
                                type: 'Number'
                            }
```

?

tzzed:
@fgalan we already send data like this 
```json
{
    "fireDetectionThreshold": 10
}
```
or by using the oid: 

```json
{
    "f_t": 10
}
```

since the version 4.0 it does not work. So after the fix of @GuillaumeOd  both works. Why did you change this functionnality? 
We really need this fix? In case of devices that use oid the iot agent does not support it anymore. 
You can see the test of @GuillaumeOd  and try it without the fix. 
Thank you again.

fgalan:
> @fgalan we already send data like this
> 
> ```json
> {
>     "fireDetectionThreshold": 10
> }
> ```
> 
> or by using the oid:
> 
> ```json
> {
>     "f_t": 10
> }
> ```
> 
> since the version 4.0 it does not work. So after the fix of @GuillaumeOd both works. Why did you change this functionnality? We really need this fix? In case of devices that use oid the iot agent does not support it anymore. You can see the test of @GuillaumeOd and try it without the fix. Thank you again.

So, you mean that you have some devices that are sending like `"f_t": 10` and *at the same time* you have some other devices that are sending like `"fireDetectionThreshold": 10`.

Is my understanding correct?

tzzed:
Yes exactly. Somes use object_id and othes use the name.

fgalan:
OK. Much clearer now :) Thanks!

> since the version 4.0 it does not work. So after the fix of @GuillaumeOd both works. Why did you change this functionnality? We really need this fix? In case of devices that use oid the iot agent does not support it anymore. You can see the test of @GuillaumeOd and try it without the fix. Thank you again.

It was not an intentional change (but not within our use cases, so it was not noticed by us).

Thus, the PR is ok, as long as:

1. Merge master on this PR's branch to check all test keep working
2. This patch on the library has to be checked with IOTA-JSON, IOTA-UL and IOTA-Manager unit test. I'll take care of it, once you complete step 1

What do you think?

tzzed:
Great thank you! that's work, @GuillaumeOd it's in your hands.

GuillaumeOd:
thanks @fgalan ! I just merged master to the branch. Please notice that the test I added was failing because of a type error. The expected is:

`fireDetectionThreshold: {
    type: 'Number',
    value: 10
}`

but for some reason we receive:

`fireDetectionThreshold: {
    type: 'Text',
    value: 10
}`

Is this the normal behavior ?

AlvaroVega:
> @AlvaroVega what do you think?

I think both branch and test were aligned.

GuillaumeOd:
> > @AlvaroVega what do you think?
> 
> I think both branch and test were aligned.

The question is, what should be the expected payload to the test I added ? It appears that we receive a type `Text` but as my knowledge is limited concerning the type question in iotagent-node-lib I don't know if it should be `Number` for this case

fgalan:
> > > @AlvaroVega what do you think?
> > 
> > 
> > I think both branch and test were aligned.
> 
> The question is, what should be the expected payload to the test I added ? It appears that we receive a type `Text` but as my knowledge is limited concerning the type question in iotagent-node-lib I don't know if it should be `Number` for this case

It should be `Number`, as it is in the current test. Thus, I guess that the code needs some extra fix to align with the tests.

GuillaumeOd:
> > > > @AlvaroVega what do you think?
> > > 
> > > 
> > > I think both branch and test were aligned.
> > 
> > 
> > The question is, what should be the expected payload to the test I added ? It appears that we receive a type `Text` but as my knowledge is limited concerning the type question in iotagent-node-lib I don't know if it should be `Number` for this case
> 
> It should be `Number`, as it is in the current test. Thus, I guess that the code needs some extra fix to align with the tests.

I agree, shall I investigate it by myself ? The link with my PR is unlikely but I can try to find the fix and add it to this PR

fgalan:
> I agree, shall I investigate it by myself ? The link with my PR is unlikely but I can try to find the fix and add it to this PR

It would be great :)

GuillaumeOd:
Hey @fgalan  @AlvaroVega  @tzzed,

The test was failing because the payload included a field named "t" that did not match any object_id defined in the device provisioning. As a result, the IotAgent treated "t" as an untyped attribute and defaulted its type to "Text".

However, the test expectation assumed "t" would be interpreted as psBatteryVoltage (with type: Number), which was provisioned with object_id: ".1.0.0.1". This mismatch caused the assertion failure.

To fix it, I updated the test payload to use the correct key ".1.0.0.1" instead of "t", so the attribute is properly mapped to psBatteryVoltage with the correct type.

Please tell me if the PR is ok for you now !

GuillaumeOd:
> > Please tell me if the PR is ok for you now !
> 
> > Hey @fgalan @AlvaroVega @tzzed,
> > The test was failing because the payload included a field named "t" that did not match any object_id defined in the device provisioning. As a result, the IotAgent treated "t" as an untyped attribute and defaulted its type to "Text".
> > However, the test expectation assumed "t" would be interpreted as psBatteryVoltage (with type: Number), which was provisioned with object_id: ".1.0.0.1". This mismatch caused the assertion failure.
> > To fix it, I updated the test payload to use the correct key ".1.0.0.1" instead of "t", so the attribute is properly mapped to psBatteryVoltage with the correct type.
> > Please tell me if the PR is ok for you now !
> 
> test 021B is still failing: https://github.com/telefonicaid/iotagent-node-lib/actions/runs/15710390480/job/44269756408?pr=1667#step:5:18568

My bad, I forgot to remove my old test. This should be fine now

test/functional/testCases.js
```diff
@@ -585,7 +598,19 @@ const testCases = [
                     },
                     psBatteryVoltage: {
                         type: 'Number',
-                        value: 23.5
+                        value: 10
```
AlvaroVega, line 601: Why not keep 23.5 instead 10 for '.1.0.0.1' and psBatteryVoltage?

GuillaumeOd: That's a copy miss when I tested these test cases https://github.com/telefonicaid/iotagent-node-lib/pull/1668/files
I rolled it back !

lib/services/ngsi/entities-NGSI-v2.js
```diff
@@ -463,7 +463,7 @@ function sendUpdateValueNgsi2(entityName, originMeasures, originTypeInformation,
             }
             //remove measures that has been shadowed by an alias (some may be left and managed later)
             //Maybe we must filter object_id if there is name == object_id
-            measures = measures.filter((item) => item.name !== currentAttr.object_id && item.name !== currentAttr.name);
+            measures = measures.filter((item) => item.name !== currentAttr.object_id);
```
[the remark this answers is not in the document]
GuillaumeOd: Done :heavy_check_mark: is that ok for you ?

CHANGES_NEXT_RELEASE
```diff
@@ -1 +1 @@
-
+- Fix attribute loss during device update when both name and object_id are defined by filtering only object_id (issue #1660)
```
```suggestion
- Fix attribute loss during device update when both name and object_id are defined by filtering only object_id (#1660)
```

PR #1668, opened 2024-11-08
add functional test case about how use special word `value`as a name of attribute 

AlvaroVega:
IMHO this test case demostrates which https://github.com/telefonicaid/iotagent-node-lib/issues/1660 is working


GuillaumeOd:
Thank for the test @AlvaroVega , this is working fine for me too. But we are sending the `attr_name` as json value, not the `object_id`.

Example:

```
json: {
    a: false,
    value: 10
}
```

And as expected this change makes the test fail for me

AlvaroVega:
Then maybe you can define your attribute with a conditional expression: 
 "expression": "value? value : t"

GuillaumeOd:
I suppose the core question from my previous comment is this: should we only be updating an attribute using its `object_id`, or should we still have the option to update it by its `name`, as was possible before this[change](https://github.com/telefonicaid/iotagent-node-lib/commit/166f7715ee612c1a583e30d2d44d123ee90f261c#diff-380bfb412c0e3663bda44b839526ff8b99f099d9eb851b904e673697f6f2a978R435)

GuillaumeOd:
The doc is still showing examples of measures sent by attribute names.

https://iotagent-node-lib.readthedocs.io/en/latest/api.html

You can find examples at "Multimeasure support" and "Measurement transformation" sections.

Please consider the test added to my [PR](https://github.com/telefonicaid/iotagent-node-lib/pull/1667/files) and reproduce it in local without my fix in order to demonstrate this.

AlvaroVega:
> The doc is still showing examples of measures sent by attribute names.
> 
> https://iotagent-node-lib.readthedocs.io/en/latest/api.html

Regarding with `Device to NGSI Mapping` section:
- object_id (optional): name of the attribute as coming from the device.
- name (mandatory): ID of the attribute in the target entity in the Context Broker. Note that id and type are not valid attribute names at Context Broker. Thus, although a measure named id or type will not break the IoT Agent, they are silently ignored and never progress toward Context Broker entities.

tzzed:
@AlvaroVega I am not sure to understand what you say. 
Do you mean that if the entity contains an attribue: 
```
temperature: {
            value: 10,
            type: 'Integer',
            creDate: 1730990416.543879,
            modDate: 1730990416.543879,
            mdNames: []
        },
```

the iotagent cannot send 
```
{
 "temperature": 42
}
```
What about real life? you mean only the `oid` works? What about the previous behavior?

AlvaroVega:
> What about real life? you mean only the `oid` works? What about the previous behavior?

Of course you can send as a measure an attribute name which will be part of entity, but then not include it in device attribute mapping.

test/functional/testCases.js
```diff
@@ -542,6 +542,20 @@ const testCases = [
...
+                            },
+                            {
+                                object_id: 't',
+                                name: 'value',
+                                type: 'Number'
+                            },
+                            {
+                                name: 'fireDetectionThreshold',
+                                type: 'Number'
+                            },
```
fgalan, line 554: What about?

```suggestion
                            {
                                name: 'fireDetectionThreshold',
                                type: 'Number'
                            },
                            {
                                object_id: 'fdt',
                                name: 'fireDetectionThreshold',
                                type: 'Number'
                            },
```

next

* Send mesuare fdt: 20 -> CB expectations fireDetectionThreshold: 20
* Send mesuare fireDetectionThreshold: 20 -> CB expectations fireDetectionThreshold: 20

tzzed: @fgalan exactly. By using the objectid and the name we should have the same result.

AlvaroVega: > What donyou mean by it does not work?

@tzzed When I applied the suggested code done by fermin the test didn't pass.

Issue #1420, opened 2023-08-25
Implement southbound flow control


mapedraza:
### Component

New feature

### Is your feature request related to a problem? Please describe

Devices that send data so ofter, or an ELT could overload the IoTA

### Describe the solution you'd like

Implement a flow control mechanism like [Context Broker](https://github.com/telefonicaid/fiware-orion/blob/19d0f95518324291c85eed87638b66a37565e25f/doc/manuals/admin/perf_tuning.md?plain=1#L204) on the IoTA

mapedraza:
> There are some options to implement flowControl in southbound:
> 
> * Without a queue: Measurements are lost if they arrive during the throttling period.
> * With a queue (bull, redis): Measurements are processed later, but none are lost.

The initial idea was the first case, analogous to how it works in Orion. The idea is that the sending system could somehow "adjust its pace." This is especially relevant in the case of automated processes that trigger the sending of thousands of measurements through IoTA. In fact, this use case could be based on the "pace" set by Orion itself, being also quite straightforward to implement

PR #1677, opened 2024-12-18
use options=flowControl when update

AlvaroVega:
related https://github.com/telefonicaid/iotagent-node-lib/issues/1420

This way measure response from iotagent relies on related ContextBroker update response

- [x] Tests should be updated
- [x] Configurable flag to enable it (enabled by default ? )
- [x] IotAgentManager: https://github.com/telefonicaid/iotagent-manager/pull/355
- [x] Doc


fgalan:
I'd suggest to make the inclusion of the flowControl option to be optional. I'd suggest to use the same approach that with another on/off features (eg. timestamp), that can be set at IOTA instance / group / device level, with the same precedence rules.

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Add useCBflowControl option to global, group and device level to force to use CB flow control (#1420)
```
```suggestion
- Add: option to force to use CB flow control with new API field useCBflowControl at group and device device level  (#1420)
- Add: config useCBflowControl (IOTA_CB_FLOW_CONTROL env var) to set CB flow control behaviour at instance level (#1420)
```

To make CNR moro uniform.

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
...
 - Add: store last measure in device (by id, apikey, service and subservice) and new API field storeLastMeasure at group and device levels (#1669)
 - Add: IOTA_STORE_LAST_MEASURE env var to set default store last measure behaviour at instance level (#1669)
```
fgalan, line 3:
```suggestion
- Add: storeLastMeasure config setting  (IOTA_STORE_LAST_MEASURE env var) to set default store last measure behaviour at instance level (#1669)
```

(Not directly related with this PR, but the CNR would look better)

doc/admin.md
```diff
@@ -444,6 +444,11 @@ For example in a device document stored in MongoDB will be extended with a subdo
 }
 ```
 
+#### `useCBflowControl`
+
+If this flag is activated, when iotAgent invokes Context Broker will use flowControl option. This flag is overwritten by
```
fgalan, line 449:
```suggestion
If this flag is activated, when iotAgent invokes Context Broker will use [flowControl option](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/perf_tuning.md#updates-flow-control-mechanism). This flag is overwritten by
```

fgalan (approved):
LGTM

PR #1683, opened 2025-01-16
set polling for autoprovisioned devices

AlvaroVega:
Autoprovisioned devices are not setting polling and transport when are created. When related group has no info about transport nither endpoint then if defaultTransport in agent is HTTP polling true is not defined in autoprovisioned device.

This bug was introduced by https://github.com/telefonicaid/iotagent-json/pull/794


fgalan (commented):
LGTM

fgalan (approved):

PR #1702, opened 2025-03-12
 Error control in JEXL transformations

GregorioBlazquez:
This PR was initially intended to add try-catch blocks to some functions, but it also includes additional test cases and suggests better error handling. Referenced Issue: #1701

Issue #1701, opened 2025-03-10
Add try catch support to JEXL toisodate transformation


mrutid:
### Component

Feature enhancement

### Is your feature request related to a problem? Please describe

    toisodate: (val) => {
                      try {
                             return new Date(val).toISOString()
                       } catch (e) {
                             return null
                       }}

Enables checking Dates with ternary operator.

AlvaroVega:
https://github.com/telefonicaid/iotagent-node-lib/blob/2088e38babd1c9ab8b9e94184055202bf6b846df/lib/jexlTranformsMap.js#L42

AlvaroVega:
And check all  other functions that could raise and exception and are not currently catched.

fgalan:
Fixed by PR #1702

PR #1703, opened 2025-03-17
Further improvements to JEXL error handling (follow-up to #1702)

GregorioBlazquez:
This PR builds upon the changes introduced in PR #1702, which addressed Issue #1701 by adding exception handling to several JEXL transformation functions. The main objective of this update is to further refine error handling and improve consistency across functions.


fgalan:
I understand CNR entry is not needed, as it is covered by the existing

> - Fix: modified JEXL transformations (toisostring, gettime, parseint, etc.) to return null instead of NaN when some unexpected situation occurs (#1701)

PR #1706, opened 2025-03-27
Fix/search group using firstly apikey and then type

AlvaroVega:


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Fix: allow match auto-provisioned devices and groups when the same type is used by several groups (which always uses different apikeys) (#1713)
```
fgalan, line 1: Should a test covering this cases be added?

AlvaroVega:
related bug was reported today by jafdiez

PR #1714, opened 2025-05-14
Task/allow headers in command

AlvaroVega:
Needed by https://github.com/telefonicaid/iotagent-json/pull/874

PR #1715, opened 2025-05-20
Update api.md

AlvaroVega:
Continuation https://github.com/telefonicaid/iotagent-node-lib/pull/1714

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -259,6 +259,7 @@ Additionally for commands (which are attributes of type `command`) the following
     transports)
 -   **headers**: extra customer headers used when send command by HTTP transport (ignored in other kinds of
     transports)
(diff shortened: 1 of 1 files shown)


doc/api.md
```diff
@@ -257,6 +257,8 @@ Additionally for commands (which are attributes of type `command`) the following
     particular IOTAs documentation for allowed values of this field in each case.
 -   **contentType**: `content-type` header used when send command by HTTP transport (ignored in other kinds of
     transports)
+-   **headers**: extra customer headers used when send command by HTTP transport (ignored in other kinds of
```
fgalan, line 260: Maybe we should explain that the value of this field is a map which keys are header names and values are jexl expressions? In addition, maybe an example could help.

AlvaroVega: Done in https://github.com/telefonicaid/iotagent-node-lib/pull/1715/commits/14dc6b09d6f560faa813122253154bc27917a8b8

fgalan (approved):
LGTM

Issue #1621, opened 2024-06-20
object_id does not work in some different format


tzzed:
### IoT Agent Node Lib version the issue has been seen with

4.5.0

### Bound or port used (API interaction)

Southbound (Devices data API)

### NGSI version

NGSIv2

### Are you running a container?

Yes, I am using a contaner (Docker, Kubernetes...)

### Image type

normal

### Expected behaviour you didn't see

The measure should be updated in orion CB. 

### Unexpected behaviour you saw

The payload is not filled.
The iotagent takes only all static_attributes when I have active attribute with `object_id: '.1.3.6.1.4.1.6302.2.1.2.1.0'` in this format.
### this fails
```bson
 {
            name: 'psBatteryVoltage',
            type: 'Number',
            object_id: '.1.0.0.1'
 }
```

###this works
```bson
 {
            name: 'psBatteryVoltage',
            type: 'Number',
            object_id: 'psBattery'
 }
```

### Steps to reproduce the problem

```sh
curl  -iX POST 'http://localhost:5004/iot/json?k=<API_KEY>&i=device001
 -d '{ 
    ".1.0.0.1": 23.5
  }'
```
in this case the payload is filled by static_attributes only. 

### Configs

```yaml
environment:
    - "IOTA_CB_HOST=orion"
    - "IOTA_CB_PORT=1026"
    - "IOTA_NORTH_PORT=4041"
    - "IOTA_REGISTRY_TYPE=mongodb"
    - "IOTA_MONGO_HOST=mongodb"
    - "IOTA_MONGO_PORT=27017"
    - "IOTA_MONGO_DB=iotagent-json"
    - "IOTA_HTTP_PORT=7896"
    - "IOTA_PROVIDER_URL=http://iot-agent:4041"
```

### Log output

```sh
time=2024-06-20T09:48:28.428Z | lvl=DEBUG | corr=n/a | trans=n/a | op=IoTAgentNGSI.Request | msg=Options: {
iot-agent  |     "url": "http://orion:1026/v2/entities?options=upsert",
iot-agent  |     "method": "POST",
iot-agent  |     "headers": {
iot-agent  |         "fiware-service": "service",
iot-agent  |         "fiware-servicepath": "/"
iot-agent  |     },
iot-agent  |     "json": {
iot-agent  |         "id": "urn:ngsild:device:4242",
iot-agent  |         "type": "device",
iot-agent  |         "project": {
iot-agent  |             "type": "Text",
[...]

PR #1721, opened 2025-07-07
Fix/attribute loss when updating a device

edwinvautier:
related to https://github.com/telefonicaid/iotagent-node-lib/issues/1660

It is the continuation of this PR https://github.com/telefonicaid/iotagent-node-lib/pull/1667 from https://github.com/GuillaumeOd as he won't work on it anymore.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,2 @@
 - Fix attribute loss during device update when both name and object_id are defined by filtering only object_id (#1660)
+- Add: hash jexl transformation
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,8 @@
+4.12.0 (Jan 28th, 2026)
+
+- Fix: set default value for defaultTransport to allow set polling and transport for autoprovisioned devices when no IOTA_DEFAULT_TRANSPORT en var is used (#1751)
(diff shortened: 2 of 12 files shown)


fgalan:
Let's see how the same test works with current master branch (to fully understand which current behaviour this PR is fixing)

fgalan:
> Or maybe the test is not being run for some reason?

This confirms that the test is run https://github.com/telefonicaid/iotagent-node-lib/actions/runs/16668949855/job/47180736642#step:5:963

imagen

edwinvautier:
Sorry for the delay, my test case did not cover the fix, should be ok now, i have set a test where object id and other attribute names conflict.

edwinvautier:
> Sorry for the delay, my test case did not cover the fix, should be ok now, i have set a test where object id and other attribute names conflict.
> 
> @fgalan should be ok now

just made a new change i missed something in my test

fgalan:
Btw, what's the relationship between this PR and issue https://github.com/telefonicaid/iotagent-node-lib/issues/1621. Does this PR solves that issue also?

fgalan (approved):
LGTM. Thanks for the contribution! :)

Maybe we do some minor adjust in CNR and code comments in a separte PR.

